PR #25, opened 2017-11-28
Config changes for RKE

galal-hussein:


apis/cluster.cattle.io/v1/types.go
```diff
@@ -120,18 +120,18 @@ type RancherKubernetesEngineConfig struct {
 	SSHKeyPath string `yaml:"ssh_key_path" json:"sshKeyPath,omitempty"`
 }
 
-type RKEConfigHost struct {
-	// SSH IP address of the host
-	IP string `yaml:"ip" json:"ip,omitempty"`
-	// Advertised address that will be used for components communication
-	AdvertiseAddress string `yaml:"advertise_address" json:"advertiseAddress,omitempty"`
-	// Host role in kubernetes cluster (controlplane, worker, or etcd)
+type RKEConfigNode struct {
```
alena1108: no real reason for renaming the type struct, but it's ok

PR #38, opened 2017-12-01
new fields for resource req and limit 

kinarashah:


apis/cluster.cattle.io/v1/types.go
```diff
@@ -219,6 +221,8 @@ type ClusterNode struct {
 	v1.Node
 	NodeName    string
 	ClusterName string
+	Requested   v1.ResourceList `json:"requested,omitempty"`
```
alena1108: Would it be possible to make it a part of the NodeSpec? Along with the Allocatable/Capacity

apis/cluster.cattle.io/v1/types.go
```diff
@@ -216,9 +218,33 @@ type AuthConfig struct {
...
-	ClusterName string
+	NodeName        string
+	ClusterName     string
+	metav1.TypeMeta `json:",inline"`
+	// Standard object’s metadata. More info:
+	// https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#metadata
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Specification of the desired behavior of the cluster node. More info:
+	// https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#spec-and-status
+	Spec v1.NodeSpec `json:"spec,omitempty"`
```
alena1108: I guess you don't need `Spec`. You can just define the type?

apis/cluster.cattle.io/v1/types.go
```diff
@@ -216,9 +218,33 @@ type AuthConfig struct {
...
+	// Most recent observed status of the cluster. More info:
+	// https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#spec-and-status
+	Status ClusterNodeStatus `json:"status"`
+}
+
+type ClusterNodeStatus struct {
+	Capacity        v1.ResourceList        `json:"capacity,omitempty"`
```
alena1108: You should embed the v1.NodeStatus, and add 2 new fields

apis/cluster.cattle.io/v1/types.go
```diff
@@ -216,9 +218,33 @@ type AuthConfig struct {
 }
 
 type ClusterNode struct {
-	v1.Node
-	NodeName    string
-	ClusterName string
+	NodeName        string
```
alena1108: can you move nodename, clustername down? so k8s native fields come first

alena1108:
@kinarashah some build failures regarding go generate, could you check? a lot of stuff seems to be missing in zz too

alena1108:
@kinarashah we should also set requested/limit on a clusterstatus in addition to clusternodestatus

kinarashah:
closing this, had a bunch of files deleting unexpectedly with go generate. #39

PR #62, opened 2017-12-14
Default project role and more RoleTemplate fields

cjellick:


ibuildthecloud:
Pretty sure this is already included

PR #63, opened 2017-12-14
add authorization mode

moelsayed:


ibuildthecloud:
I think I'd prefer we rename `auth` to `authentication`.  I know it's a breaking change, but RKE is still really in a "beta" status.

PR #64, opened 2017-12-14
Changes to Token type to add explicit Identity fields

prachidamle:
Changing token type to have explicit fields UserIdentity, GroupIdentities and ProviderInfo


cjellick:
LGTM

PR #65, opened 2017-12-14
serviceAccountSecretName

kinarashah:


alena1108:
Closing; will open a new one if there is still a need for this feature.

PR #90, opened 2017-12-26
add core to management api

StrongMonkey:


StrongMonkey:
already included in https://github.com/rancher/types/pull/91.

PR #95, opened 2017-12-29
Namespace things

cjellick:
This is likely going to break a bunch of stuff until we get the automagic namespacing working

Also, the bug/quirk with maps during code generation casuing random changes should be fixed and `go generate` reran before merging.


ibuildthecloud:
Included in #97

PR #103, opened 2018-01-03
Add secrets

StrongMonkey:


StrongMonkey:
This pr adds secret type and helm release info to namespace.

PR #139, opened 2018-01-22
Use 'dnsLabel' for Cluster name and machine requestedHostname

moelsayed:
https://github.com/rancher/rancher/issues/10965
https://github.com/rancher/management-api/pull/35

### DIFF
--- a/apis/management.cattle.io/v3/schema/schema.go
+++ b/apis/management.cattle.io/v3/schema/schema.go
@@ -67,7 +67,15 @@ func clusterTypes(schemas *types.Schemas) *types.Schemas {
 			schema.SubContext = "clusters"
 		}).
 		MustImport(&Version, v3.ClusterEvent{}).
-		MustImport(&Version, v3.ClusterRegistrationToken{})
+		MustImport(&Version, v3.ClusterRegistrationToken{}).
+		MustImportAndCustomize(&Version, v3.Cluster{}, func(schema *types.Schema) {
+			schema.MustCustomizeField("name", func(field types.Field) types.Field {
+				field.Type = "dnsLabel"
+				field.Nullable = true
+				field.Required = false
+				return field
+			})
+		})
 }
 
 func authzTypes(schemas *types.Schemas) *types.Schemas {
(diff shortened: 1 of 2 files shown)


apis/management.cattle.io/v3/schema/schema.go
```diff
@@ -66,7 +66,15 @@ func clusterTypes(schemas *types.Schemas) *types.Schemas {
 			schema.SubContext = "clusters"
 		}).
 		MustImport(&Version, v3.ClusterEvent{}).
-		MustImport(&Version, v3.ClusterRegistrationToken{})
+		MustImport(&Version, v3.ClusterRegistrationToken{}).
+		MustImportAndCustomize(&Version, v3.Cluster{}, func(schema *types.Schema) {
+			schema.MustCustomizeField("name", func(field types.Field) types.Field {
+				field.Type = "dnsLabel"
```
alena1108: any chance `dnsLabel` is a constant in norman? so we can use that constant instead of hardcoding the naming

moelsayed:
@ibuildthecloud One thing I noticed. After using the `dnsLabel` type, the field stopped showing as `Nullable` in the API. I couldn't figure out why, and I am not sure if that's a problem.

apis/management.cattle.io/v3/schema/schema.go
```diff
@@ -183,7 +191,24 @@ func machineTypes(schemas *types.Schemas) *types.Schemas {
 				Output: "machineDriver",
 			}
 		}).
-		MustImport(&Version, v3.MachineTemplate{})
+		MustImport(&Version, v3.MachineTemplate{}).
+		MustImportAndCustomize(&Version, v3.Machine{}, func(schema *types.Schema) {
+			schema.MustCustomizeField("requestedHostname", func(field types.Field) types.Field {
```
ibuildthecloud: For types that we define (not a k8s native type) we can just annotate the struct itself and not modify the schema "after the fact."  To do this for "requestedHostname" it would be in the Cluster struct

```go
type Cluster struct [
    RequestedHostname string `json:"requestedHostname,omitempty" norman:"type=dnsLabel,nullable,required"`
}
```

The approach does not work for the `name` field because we do define that field and it comes from ObjectMeta.  So for "name" you must use ImportAndCustomize approach

alena1108:
@moelsayed looks like drone is failing, most likely generated changes are missing?

PR #156, opened 2018-01-29
PublicEndpoints to pod/node/service

alena1108:


apis/project.cattle.io/v3/types.go
```diff
@@ -141,3 +141,16 @@ type SSHAuth struct {
 	Fingerprint string `json:"certFingerprint" norman:"nocreate,noupdate"`
 }
 type NamespacedSSHAuth SSHAuth
+
+type PublicEndpoint struct {
+	NodeName  string `json:"nodeName,omitempty" norman:"nocreate,noupdate"`
```
ibuildthecloud: Can this be `type=reference[node]`?

apis/project.cattle.io/v3/types.go
```diff
@@ -141,3 +141,16 @@ type SSHAuth struct {
...
+	NodeIP    string `json:"nodeIp,omitempty" norman:"nocreate,noupdate"`
+	Port      int32  `json:"port,omitempty" norman:"nocreate,noupdate"`
+	Protocol  string `json:"protocol,omitempty" norman:"nocreate,noupdate"`
+	Namespace string `json:"namespace,omitempty" norman:"nocreate,noupdate"`
+	// for node port service
+	Service string `json:"service,omitempty" norman:"nocreate,noupdate"`
+	// for host port
+	Pod string `json:"pod" norman,omitempty:"nocreate,noupdate"`
```
ibuildthecloud: Can this be `type=reference[pod]`?  That would mean that the value would be "namespace:podName" format. The same with service.  Then there is no need for a namespace param.

PR #159, opened 2018-01-31
Add ingress config to rke and labels to nodes

galal-hussein:


apis/management.cattle.io/v3/rke_types.go
```diff
@@ -164,3 +168,12 @@ type AuthzConfig struct {
...
 	Options map[string]string `yaml:"options" json:"options,omitempty"`
 }
+
+type IngressConfig struct {
+	// Ingress controller type used by kubernetes
+	Type string `yaml:"type" json:"type,omitempty"`
+	// Ingress controller options
+	Options map[string]string `yaml:"options" json:"options,omitempty"`
+	// NodeSelector key pair
+	NodeSelector map[string]string `yaml:"node_selector" json:"nodeSelector,omitempty"`
```
alena1108: I'd recommend NodeSelector to be in Ingress config options - we can document the key. So in case we want some other deployment model in the future - like adding parameter for taints, we can keep it generic

PR #180, opened 2018-02-06
update logging related fields

aiwantaozi:
Related issue: https://github.com/rancher/rancher/issues/11167
changes:
1. replace the host, port with endpoint
2. use norman tag to add field default, enum value


aiwantaozi:
@cjellick @ibuildthecloud could you help review the logging fields changes.

apis/management.cattle.io/v3/logging_types.go
```diff
@@ -87,47 +87,35 @@ type LoggingCondition struct {
...
-
-type Zookeeper struct {
-	Host string `json:"host,omitempty"`
-	Port int    `json:"port,omitempty"`
-}
-
-type BrokerList struct {
-	BrokerList []string `json:"brokerList,omitempty"`
+	ZookeeperEndpoint string `json:"zookeeperEndpoint,omitempty"`
+	BrokerEndpoints   string `json:"brokerEndpoints,omitempty"`
```
alena1108: there can be multiple endpoints for brokerEndpoints? if so, any reason why we don't use []string type?

apis/management.cattle.io/v3/logging_types.go
```diff
@@ -87,47 +87,35 @@ type LoggingCondition struct {
...
+	DateFormat   string `json:"dateFormat,omitempty" norman:"required,type=enum,options=YYYY-MM-DD|YYYY-MM|YYYY,default=YYYY-MM-DD"`
 	AuthUserName string `json:"authUsername,omitempty"` //secret
 	AuthPassword string `json:"authPassword,omitempty"` //secret
 }
 
 type SplunkConfig struct {
-	Host     string `json:"host,omitempty"`
-	Port     int    `json:"port,omitempty"`
-	Protocol string `json:"protocol,omitempty"`
+	Endpoint string `json:"endpoint,omitempty" norman:"required"`
```
alena1108: What is the format for the endpoint? Is it meant as a replacement for address/port/protocol? If yes, is there any reason why we don't use a composite field for that, something similar to:

https://github.com/rancher/types/blob/d6b26cedb92786323fb79adf4af529d927366fd7/apis/cluster.cattle.io/v3/schema/schema.go#L71

aiwantaozi: I see in the 1.6 the publicEndpoint like this, still the same?
```
"publicEndpoints": [ 
{
"type": "publicEndpoint",
"hostId": "1h7",
"instanceId": "1i510503",
"ipAddress": "120.76.202.180",
"port": 8081,
"serviceId": "1s629"
},
{
"type": "publicEndpoint",
"hostId": "1h7",
"instanceId": "1i526194",
"ipAddress": "120.76.202.180",
"port": 500,
"serviceId": "1s5"
}]
```

current the public endpoint for different target I expected user may input different format, for example: elasticsearch: https://ip:port, syslog: ip:port.(because it use udp/tcp protocol)

apis/management.cattle.io/v3/logging_types.go
aiwantaozi: it can be multiple endpoints for brokerEndpoints, I can change to use []string.

apis/management.cattle.io/v3/logging_types.go
alena1108: @aiwantaozi thats fine, spoke with @vincent99, from the UI perspective we are good having endpoint as a string.

alena1108:
@aiwantaozi could you split the PR into 2 commits: first one introducing changes in types files, second - with generated changes. Thats the format that we follow so it is easier to resolve merge conflicts.

PR #191, opened 2018-02-08
add download url as required

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/11132

PR #201, opened 2018-02-11
Add types for pipeline

gitlawr:


apis/management.cattle.io/v3/pipeline_types.go
```diff
@@ -0,0 +1,224 @@
...
+
+	PipelineExecutionName string `json:"pipelineExecutionName,omitempty" norman:"type=reference[pipelineExecution]"`
+	Stage                 int    `json:"stage,omitempty" norman:"min=1"`
+	Step                  int    `json:"step,omitempty" norman:"min=1"`
+	Line                  int    `json:"line,omitempty"`
+	Message               string `json:"message,omitempty"`
+}
+
+type AuthAppInput struct {
+	Type         string `json:"type,omitempty" norman:"type=string,required"`
```
alena1108: I recall type having special meaning in norman API, and we should avoid this name as a field. We've had some problem using Type field in `ingress` embedded types, so it had to be fixed: https://github.com/rancher/types/pull/185/commits/c00024288496c10960dc0bcbbcd7581ac097fa56. Can you double check if it won't result in any problems in your case.

gitlawr: The Input object can be parsed in actionHandler apart from the framework. But I think it's a good idea to avoid using the type field to keep consistent with other places. I've updated the commit.

gitlawr:
Rebased.

orangedeng:
Related Issue https://github.com/rancher/rancher/issues/11400

PR #206, opened 2018-02-12
Plan types

moelsayed:
https://github.com/rancher/rke/pull/335


apis/management.cattle.io/v3/rke_types.go
```diff
@@ -203,3 +205,40 @@ type IngressConfig struct {
 	// NodeSelector key pair
 	NodeSelector map[string]string `yaml:"node_selector" json:"nodeSelector,omitempty"`
 }
+
+type RKEPlan struct {
+	// List of node Plans
+	Nodes []RKEConfigNodePlan `json:"configNodePlan,omitempty"`
```
alena1108: any reason why json tag is diff from param name? we usually align it.

PR #221, opened 2018-02-15
K8s version SystemImages Defaults

moelsayed:
- Move k8s default SystemImages map to types

https://github.com/rancher/rancher/issues/11363


apis/management.cattle.io/v3/k8s_defaults.go
```diff
@@ -0,0 +1,59 @@
+package v3
+
+var (
+	K8sVersionToSystemImages = map[string]RKESystemImages{
```
alena1108: @moelsayed @galal-hussein Looks good, but lets think of other naming to better reflect the purpose. `K8sVersionToRKESystemImages` and `RancherK8sVersionToSystemImages` perhaps?

PR #233, opened 2018-02-20
Compose support

StrongMonkey:


StrongMonkey:
need https://github.com/rancher/norman/pull/91 change to add yaml tags

PR #283, opened 2018-03-06
Add SSHAgentAuth boolean

cloudnautique:
This is added to both the global and Node config level


alena1108:
@cloudnautique could you split it to 2 commits - commit 1 with changes to rke_types.go, and the second commit with generated changes

PR #284, opened 2018-03-06
Change sourceCode resources to user scoped

gitlawr:
Should coordinate with rancher PR: https://github.com/rancher/rancher/pull/11989
Related to https://github.com/rancher/rancher/issues/11977
1. change sourceCodeCredential/Repository to user-scoped so users won't have access to other users account/repositories.
2. Because Pipelines are in project-scoped. When user A sees user B's pipelines we need some info of B's Github account to show(like avatar, Github username, etc.), so we store a sourcecodecredential reference in pipelinestatus with sensitive data erased.


apis/management.cattle.io/v3/pipeline_types.go
```diff
@@ -78,14 +82,15 @@ type GithubClusterConfig struct {
...
-	Token           string `json:"token,omitempty" yaml:"token,omitempty" norman:"writeOnly,noupdate"`
+	PipelineState        string                `json:"pipelineState,omitempty" norman:"required,options=active|inactive,default=active"`
+	NextRun              int                   `json:"nextRun" yaml:"nextRun,omitempty" norman:"default=1,min=1"`
+	LastExecutionID      string                `json:"lastExecutionId,omitempty" yaml:"lastExecutionId,omitempty"`
+	LastRunState         string                `json:"lastRunState,omitempty" yaml:"lastRunState,omitempty"`
+	LastStarted          string                `json:"lastStarted,omitempty" yaml:"lastStarted,omitempty"`
+	NextStart            string                `json:"nextStart,omitempty" yaml:"nextStart,omitempty"`
+	WebHookID            string                `json:"webhookId,omitempty" yaml:"webhookId,omitempty"`
+	Token                string                `json:"token,omitempty" yaml:"token,omitempty" norman:"writeOnly,noupdate"`
+	SourceCodeCredential *SourceCodeCredential `json:"sourceCodeCredential,omitempty" yaml:"sourceCodeCredential,omitempty"`
```
zionwu: Why not use a id here?  and set the field `norman:"type:reference[sourceCodeCredential]"`

gitlawr: As stated in the PR description:
>Because Pipelines are in project-scoped. When user A sees user B's pipelines we need some info of B's Github account to show(like avatar, Github username, etc.), so we store a sourcecodecredential reference in pipelinestatus with sensitive data erased.

zionwu: ok, got it.  If the sourceCodeCredential is changed, will update the status here?

apis/management.cattle.io/v3/schema/schema.go
```diff
@@ -429,7 +424,15 @@ func pipelineTypes(schema *types.Schemas) *types.Schemas {
 				"rerun": {},
 			}
 		}).
-		MustImport(&Version, v3.SourceCodeRepository{}).
-		MustImport(&Version, v3.PipelineExecutionLog{})
+		MustImport(&Version, v3.PipelineExecutionLog{}).
+		MustImportAndCustomize(&Version, v3.SourceCodeCredential{}, func(schema *types.Schema) {
```
zionwu: if is is not allowed to be created via API, we should probably remove the POST method from collectionMethods.

apis/management.cattle.io/v3/pipeline_types.go
gitlawr: It should, but we don't expect it to be changeable by users.

apis/management.cattle.io/v3/schema/schema.go
gitlawr: Sure, updated

zionwu:
LGTM

alena1108:
LGTM

PR #288, opened 2018-03-14
Add Cluster Name for KubeConfig

HighwayofLife:
Need this for [rke PR 413](https://github.com/rancher/rke/pull/413)


galal-hussein:
Thanks for the PR, you should run go generate after adding this field to the structure, that's why builds are failing

HighwayofLife:
Thanks, I've generated changes. Rebased and updated the PR.

PR #293, opened 2018-03-16
Pipeline enhancements

gitlawr:
1. Add pipeline conditions fields, for https://github.com/rancher/rancher/issues/11983.
2. Add support for pull request trigger, for https://github.com/rancher/rancher/issues/11678.
3. Add default branch field in repo resource.


apis/management.cattle.io/v3/pipeline_types.go
```diff
@@ -1,10 +1,19 @@
...
 	"github.com/rancher/norman/types"
+	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type PipelineConditionType string
+
+const (
+	PipelineExecutionConditonProvisioned condition.Cond = "Provisioned"
+	PipelineExecutionConditionFailed     condition.Cond = "Failed"
```
alena1108: this is a bit vague for a name. Any condition can fail, I would not recommend to use it for the name. Plus condition should be "positive", meaning that when it's value is true, everything is good. Otherwise state aggregation on api level won't work by default.

What does this condition mean? Is it an equivalent of Provisioned=false? Or is it more about execution itself? If the latter, can we change it to be `PipelineExecutionConditionUpdated` or `PipelineExecutionConditionExecuted`

gitlawr: It means the pipelineexecution failed the build, for example got non-zero exit of the scripts. So is it good to be ConditionSucceed, updated/executed is not what it supposed to be.

alena1108: @gitlawr PipelineExecutionConditionSucceed or Finished is better, as it falls under rule that when set to `True`, it's a good thing

alena1108:
LGTM

PR #294, opened 2018-03-16
Psp permissions

nathan-jenan-rancher:
Adding a pod security policy template project binding resource type so
that separate permissions between projects and editing their pspts can be
established.

Issue:
https://github.com/rancher/rancher/issues/12160

Dependency of:
https://github.com/rancher/rancher/pull/12141


apis/management.cattle.io/v3/authz_types.go
```diff
@@ -91,6 +90,15 @@ type PodSecurityPolicyTemplate struct {
 	Spec        extv1.PodSecurityPolicySpec `json:"spec,omitempty"`
 }
 
+type PodSecurityPolicyTemplateProjectBinding struct {
+	types.Namespaced
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	PodSecurityPolicyTemplateID string `json:"podSecurityPolicyTemplate" norman:"required"`
```
cjellick: To follow the established pattern, the name of the field should be `PodSecurityPolicyTemplateName` the json field shoudl be `PodSecurityPolicyTemplateId` and it should have `type=reference[podSecurityPolicyTemplate]` in the norman tag

apis/management.cattle.io/v3/authz_types.go
```diff
@@ -91,6 +90,15 @@ type PodSecurityPolicyTemplate struct {
...
+	ProjectID                   string `json:"projectId" norman:"required"`
```
cjellick: same as above

apis/management.cattle.io/v3/authz_types.go
```diff
@@ -45,10 +45,9 @@ type ProjectCondition struct {
 }
 
 type ProjectSpec struct {
-	DisplayName                   string `json:"displayName,omitempty" norman:"required"`
-	Description                   string `json:"description"`
-	ClusterName                   string `json:"clusterName,omitempty" norman:"required,type=reference[cluster]"`
-	PodSecurityPolicyTemplateName string `json:"podSecurityPolicyTemplateName,omitempty" norman:"type=reference[podSecurityPolicyTemplate]"`
+	DisplayName string `json:"displayName,omitempty" norman:"required"`
+	Description string `json:"description"`
+	ClusterName string `json:"clusterName,omitempty" norman:"required,type=reference[cluster]"`
```
cjellick: where did the pspt field go? you should be adding that to projectStatus, correct?

apis/management.cattle.io/v3/schema/schema.go
```diff
@@ -149,14 +149,26 @@ func authzTypes(schemas *types.Schemas) *types.Schemas {
...
+					Input:  "setPodSecurityPolicyTemplateInput",
+					Output: "project",
+				},
+			}
+		}).
 		MustImport(&Version, v3.GlobalRole{}).
 		MustImport(&Version, v3.GlobalRoleBinding{}).
 		MustImport(&Version, v3.RoleTemplate{}).
 		MustImport(&Version, v3.PodSecurityPolicyTemplate{}).
+		MustImport(&Version, v3.PodSecurityPolicyTemplateProjectBinding{}).
```
cjellick: Right now, the PSPTB resource is probably exposed as an api endpoint at /v3 (you should be able to see it in your browser). This also means that by default, CRUD operations will be exposed directly in the API. We want to prevent that. I think we can do so by customizing the schema and setting CollectionMethods and ResourceMethods to empty slices. (These default to values like POST, GET, PUT, DELETE)

apis/management.cattle.io/v3/authz_types.go
```diff
@@ -116,3 +124,7 @@ type ClusterRoleTemplateBinding struct {
 	ClusterName        string `json:"clusterName,omitempty" norman:"required,type=reference[cluster]"`
 	RoleTemplateName   string `json:"roleTemplateName,omitempty" norman:"required,type=reference[roleTemplate]"`
 }
+
+type SetPodSecurityPolicyTemplateInput struct {
+	PodSecurityPolicyTemplate string `json:"podSecurityPolicyTemplate" norman:"required"`
```
cjellick: follow saming naming paradign as above

cjellick: regarding this comment:
> follow saming naming paradign as above
Note that once you have it setup properly, the UI should display a drop down of pspts, not a text box
psptaction

cjellick: and testing this has led be to discover that the PSPT is misconfigured. It should have a displayName field and the displayName mapper.
Right now, when i create one, the name given to it is used as the ID. This is incorrect. The id should be autogenerated and their shoudl be a separate API name field with the name the user entered. This api name field should be stored in K8S as displayName (to avoid naming collision with objectmeta.Name, which k8s uses as the id).

When fixed, the url for a pspt should looke like this:
```
https://localhost:8443/v3/podSecurityPolicyTemplates/<semi random thing that will look like podsecuritypolicytempalte-kq3jh>
```
not this (which is what it is currently):
```
https://localhost:8443/v3/podSecurityPolicyTemplates/test1
```

apis/management.cattle.io/v3/authz_types.go
nathan-jenan-rancher: Added status to: https://github.com/rancher/types/pull/294/files#diff-2e11c852ab2dcd25c29b83183a648fa3R30

cjellick:
LGTM

PR #297, opened 2018-03-19
add the node status updated to support update custom node role

aiwantaozi:
add updated state to support update role in add custom node


zionwu:
https://github.com/rancher/rancher/issues/11341

PR #305, opened 2018-03-22
Add alt names

HighwayofLife:
Adding support for additional DNS names and IPs (Alt names) used for PKI (certificates) in rke.

### DIFF
--- a/apis/management.cattle.io/v3/rke_types.go
+++ b/apis/management.cattle.io/v3/rke_types.go
@@ -237,6 +237,8 @@ type AuthnConfig struct {
 	Strategy string `yaml:"strategy" json:"strategy,omitempty"`
 	// Authentication options
 	Options map[string]string `yaml:"options" json:"options,omitempty"`
+	// List of additional hostnames and IPs to include in the api server PKI cert
+	SANs []string `yaml:"sans" json:"sans,omitempty"`
 }
 
 type AuthzConfig struct {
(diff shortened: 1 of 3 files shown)


apis/management.cattle.io/v3/rke_types.go
```diff
@@ -237,6 +237,8 @@ type AuthnConfig struct {
 	Strategy string `yaml:"strategy" json:"strategy,omitempty"`
 	// Authentication options
 	Options map[string]string `yaml:"options" json:"options,omitempty"`
+	// List of additional hostnames and IPs to include in the api server PKI cert
+	Sans []string `yaml:"sans" json:"sans,omitempty"`
```
ibuildthecloud: nit: This should probably be `SANs` not `sans` as SAN is an acronym.

HighwayofLife: Which part? -- Can't be the yaml, as we have nothing else that has any capitalization in cluster.yml. Maybe the variable?

ibuildthecloud: Sorry, yes the variable.  Just to follow go naming conventions.  Go capitalizes acronyms.

deniseschannon:
@HighwayofLife could you fix based on @ibuildthecloud comments?

HighwayofLife:
@deniseschannon done, go-generated and rebased. -- waiting for CI build to complete now.

PR #320, opened 2018-03-28
Update rke service sidekick default image

galal-hussein:
waiting until a new image of rke service sidekick is pushed


galal-hussein:
rancher/rke-service-sidekick:v0.1.1 was pushed, @alena1108 can you please review?

PR #329, opened 2018-03-30
Embedded res quota

aiwantaozi:
Related issue： https://github.com/rancher/rancher/issues/11779
Add resource quota to embedded logging

### DIFF
--- a/apis/management.cattle.io/v3/logging_types.go
+++ b/apis/management.cattle.io/v3/logging_types.go
@@ -105,6 +105,10 @@ type EmbeddedConfig struct {
 	DateFormat            string `json:"dateFormat,omitempty" norman:"required,type=enum,options=YYYY-MM-DD|YYYY-MM|YYYY,default=YYYY-MM-DD"`
 	ElasticsearchEndpoint string `json:"elasticsearchEndpoint,omitempty" norman:"nocreate"`
 	KibanaEndpoint        string `json:"kibanaEndpoint,omitempty" norman:"nocreate"`
+	RequestsMemery        int    `json:"requestsMemory,omitempty" norman:"default=500,min=500"`
+	RequestsCPU           int    `json:"requestsCpu,omitempty" norman:"default=1000,min=1000"`
+	LimitsMemery          int    `json:"limitsMemory,omitempty"`
+	LimitsCPU             int    `json:"limitsCpu,omitempty"`
 }
 
 type KafkaConfig struct {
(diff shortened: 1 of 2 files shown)


apis/management.cattle.io/v3/logging_types.go
```diff
@@ -105,6 +105,10 @@ type EmbeddedConfig struct {
 	DateFormat            string `json:"dateFormat,omitempty" norman:"required,type=enum,options=YYYY-MM-DD|YYYY-MM|YYYY,default=YYYY-MM-DD"`
 	ElasticsearchEndpoint string `json:"elasticsearchEndpoint,omitempty" norman:"nocreate"`
 	KibanaEndpoint        string `json:"kibanaEndpoint,omitempty" norman:"nocreate"`
+	RequestsMemery        int    `json:"requestsMemory,omitempty" norman:"default=500,min=500"`
+	RequestsCPU           int    `json:"requestsCPU,omitempty" norman:"default=1000,min=1000"`
+	LimitsMemery          int    `json:"limitsMemory,omitempty"`
```
gitlawr: Do we need `min=xxx` fot the limit fields

aiwantaozi: the limits is not required, if add min here, it must has value, so I just add the validation in the api. https://sourcegraph.com/github.com/aiwantaozi/rancher@e1d6c63607eaf6e1999ed572b757e5c7c5d753ae/-/blob/pkg/controllers/user/logging/utils/wrap.go#L221

apis/management.cattle.io/v3/logging_types.go
```diff
@@ -105,6 +105,10 @@ type EmbeddedConfig struct {
 	DateFormat            string `json:"dateFormat,omitempty" norman:"required,type=enum,options=YYYY-MM-DD|YYYY-MM|YYYY,default=YYYY-MM-DD"`
 	ElasticsearchEndpoint string `json:"elasticsearchEndpoint,omitempty" norman:"nocreate"`
 	KibanaEndpoint        string `json:"kibanaEndpoint,omitempty" norman:"nocreate"`
+	RequestsMemery        int    `json:"requestsMemory,omitempty" norman:"default=500,min=500"`
+	RequestsCPU           int    `json:"requestsCPU,omitempty" norman:"default=1000,min=1000"`
```
alena1108: we should not cap in json tags; caps should be there only in golang vars. It shoudl be just requestCpu in json (you can check other *cpu vars under types, it follows this paradigm)

PR #331, opened 2018-03-30
Change to k8s 1.8.10

superseb:


alena1108:
the fix will be incorporated to https://github.com/rancher/types/pull/332

PR #340, opened 2018-04-02
fixes for deployment actions  

kinarashah:
https://github.com/rancher/rancher/issues/11692#issuecomment-377613655


apis/project.cattle.io/v3/types.go
```diff
@@ -107,3 +107,7 @@ type Workload struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
 }
+
+type DeploymentRollbackInput struct {
+	ReplicaSetID string `json:"replicaSetId" norman:"type=reference[/v3/schemas/replicaSet]"`
```
alena1108: I don't think you need v3/schemas. That is only needed when referenced from some other package - like management

PR #359, opened 2018-04-10
Add json annotation for cloud provider and iface

galal-hussein:
https://github.com/rancher/rancher/issues/11780


galal-hussein:
@loganhz ^

PR #364, opened 2018-04-11
[WIP]update default resource request

aiwantaozi:
set embedded elasticsearch memory and cpu higher, make the elasticsearch will not easily get OOM at the beginning.


zionwu:
LGTM

PR #368, opened 2018-04-15
 Renamed ExtraArguments to ExtraArgs at IngressConfig rke types

rawmind0:
Needed for https://github.com/rancher/rke/pull/501


alena1108:
@rawmind0 LGTM, but can you put a brief explanation in the future why we are doing certain changes? like for readability, or if there is a bug, etc...

PR #374, opened 2018-04-17
User major version for kubernetes services options

moelsayed:
https://github.com/rancher/rke/issues/511


apis/management.cattle.io/v3/k8s_defaults.go
```diff
@@ -153,3 +157,13 @@ var (
 		},
 	}
 )
+
+func GetKubernetesServicesOptions(version string) KubernetesServicesOptions {
```
alena1108: do you think we could define this function in RKE? lets keep type for object definition, not helper functions

PR #376, opened 2018-04-18
fields for git tag trigger support in pipeline

gitlawr:
Add `TriggerWebhookTag` field
Add `OptionalEnvVars ` field to store environment variables which is not in execution fields. It stores `CICD_GIT_TAG` for now and is open for extension.

Related issue: https://github.com/rancher/rancher/issues/12624


apis/management.cattle.io/v3/pipeline_types.go
```diff
@@ -181,11 +182,12 @@ type PipelineExecutionSpec struct {
...
-	ExecutionState string        `json:"executionState,omitempty"`
-	Started        string        `json:"started,omitempty"`
-	Ended          string        `json:"ended,omitempty"`
-	Stages         []StageStatus `json:"stages,omitempty"`
+	Commit          string            `json:"commit,omitempty"`
+	ExecutionState  string            `json:"executionState,omitempty"`
+	Started         string            `json:"started,omitempty"`
+	Ended           string            `json:"ended,omitempty"`
+	Stages          []StageStatus     `json:"stages,omitempty"`
+	OptionalEnvVars map[string]string `json:"optionalEnvVars,omitempty"`
```
alena1108: having *optional* in the name doesn't sound quite right. Lets consider calling it `envVars` or something like that

PR #390, opened 2018-04-23
add gke httplb to true by default

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/12996

PR #391, opened 2018-04-24
default targetkey to sourcekey

kinarashah:
https://github.com/rancher/rancher/issues/12779

### DIFF
--- a/mapper/env.go
+++ b/mapper/env.go
@@ -112,7 +112,8 @@ func (e EnvironmentMapper) ToInternal(data map[string]interface{}) {
 		}
 
 		targetKey := convert.ToString(value["targetKey"])
-		if targetKey == "" {
+		sourceKey := convert.ToString(value["sourceKey"])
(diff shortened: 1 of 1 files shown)


mapper/env.go
```diff
@@ -132,9 +133,12 @@ func (e EnvironmentMapper) ToInternal(data map[string]interface{}) {
 				})
 			}
 		} else {
+			if targetKey == "" {
+				targetKey = sourceKey
+			}
 			switch source {
 			case "field":
-				envVar = append(envVarFrom, map[string]interface{}{
+				envVar = append(envVar, map[string]interface{}{
```
kinarashah: This is a separate bug - need to append to `envVar` not `envVarFrom`.

alena1108: @kinarashah can we link this bug to PR as well? and if not there, we should file it so qa can verify it

alena1108:
LGTM

PR #398, opened 2018-04-25
use specific pipeline image tags

gitlawr:
Related images need to be pushed to rancher registries such as `https://hub.docker.com/r/rancher/jenkins-jenkins`
@deniseschannon


deniseschannon:
LGTM

PR #403, opened 2018-04-25
add status mapper

aiwantaozi:
https://github.com/rancher/rancher/issues/13039
Flat status


aiwantaozi:
Will update to fix the build failed problem

PR #413, opened 2018-04-30
Pluggable driver

nathan-jenan-rancher:
This change gives rancher the ability to accept pluggable kontainer engine drivers
by adding the necessary types and removing the driver specific types.
    
Issue:
rancher/rancher#12698


apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -120,110 +122,31 @@ type ClusterCondition struct {
...
-	// Network
-	Network string `json:"network,omitempty"`
-	// Sub Network
-	SubNetwork string `json:"subNetwork,omitempty"`
-	// Configuration for LegacyAbac
-	EnableLegacyAbac            bool   `json:"enableLegacyAbac,omitempty"`
-	EnableStackdriverLogging    *bool  `json:"enableStackdriverLogging,omitempty" norman:"default=true"`
-	EnableStackdriverMonitoring *bool  `json:"enableStackdriverMonitoring,omitempty" norman:"default=true"`
-	MaintenanceWindow           string `json:"maintenanceWindow"`
+func init() {
```
cjellick: move to top of file

StrongMonkey:
LGTM

apis/management.cattle.io/v3/schema/schema.go
```diff
@@ -140,14 +140,6 @@ func clusterTypes(schemas *types.Schemas) *types.Schemas {
 		MustImport(&Version, v3.ImportClusterYamlInput{}).
 		MustImport(&Version, v3.ImportYamlOutput{}).
 		MustImport(&Version, v3.ExportOutput{}).
-		MustImportAndCustomize(&Version, v3.ETCDService{}, func(schema *types.Schema) {
```
StrongMonkey: hmm...did you accidentally remove this?

apis/management.cattle.io/v3/schema/schema.go
```diff
@@ -253,7 +245,6 @@ func nodeTypes(schemas *types.Schemas) *types.Schemas {
 			schema.ResourceFields["clusterId"] = clusterField
 			schema.ResourceActions["cordon"] = types.Action{}
 			schema.ResourceActions["uncordon"] = types.Action{}
-			schema.ResourceActions["stopDrain"] = types.Action{}
```
StrongMonkey: is this related to your change?

apis/management.cattle.io/v3/kontainer_types.go
```diff
@@ -0,0 +1,42 @@
...
+	// Most recent observed status of the cluster. More info:
+	// https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#spec-and-status
+	Status KontainerDriverStatus `json:"status"`
+}
+
+type KontainerDriverStatus struct {
+	ActualURL      string      `json:"actualUrl"`
+	ExecutablePath string      `json:"executablePath"`
+	Conditions     []Condition `json:"conditions"`
+	DisplayName    string      `json:"displayName"`
```
StrongMonkey: should displayName live in spec? i think that's how mapper works to find that field...

nathan-jenan-rancher: It should not live in spec because it is a computed field, not settable by the user.  Also, the mapping is working correctly even with it in status instead of spec.

apis/management.cattle.io/v3/schema/schema.go
nathan-jenan-rancher: Yes, I did.  Fixing now.

apis/management.cattle.io/v3/schema/schema.go
```diff
@@ -578,3 +578,17 @@ func clusterCatalogTypes(schemas *types.Schemas) *types.Schemas {
 			}
 		})
 }
+
+func kontainerTypes(schemas *types.Schemas) *types.Schemas {
+	return schemas.
+		AddMapperForType(&Version, v3.KontainerDriver{},
```
cjellick: drop the generic config field

client/management/v3/zz_generated_kontainer_driver.go
```diff
@@ -0,0 +1,137 @@
...
+	KontainerDriverFieldUUID                 = "uuid"
+	KontainerDriverFieldWhitelistDomains     = "whitelistDomains"
+)
+
+type KontainerDriver struct {
+	types.Resource
+	Active               bool              `json:"active,omitempty" yaml:"active,omitempty"`
+	ActualURL            string            `json:"actualUrl,omitempty" yaml:"actualUrl,omitempty"`
+	Annotations          map[string]string `json:"annotations,omitempty" yaml:"annotations,omitempty"`
+	BuiltIn              bool              `json:"builtIn,omitempty" yaml:"builtIn,omitempty"`
```
cjellick: make either readonly or drop it completely

apis/management.cattle.io/v3/schema_types.go
```diff
@@ -51,6 +52,7 @@ type Field struct {
 	ValidChars   string   `json:"validChars,omitempty"`
 	InvalidChars string   `json:"invalidChars,omitempty"`
 	Description  string   `json:"description,omitempty"`
+	DynamicField bool     `json:"dynamicField,omitempty"`
```
ibuildthecloud: What is DynamicField used for?

apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -61,19 +70,17 @@ type Cluster struct {
...
-	DockerRootDir                        string                               `json:"dockerRootDir,omitempty" norman:"default=/var/lib/docker"`
-	EnableNetworkPolicy                  *bool                                `json:"enableNetworkPolicy" norman:"default=false"`
+	DisplayName                          string                         `json:"displayName" norman:"required"`
+	Description                          string                         `json:"description"`
+	Internal                             bool                           `json:"internal" norman:"nocreate,noupdate"`
+	DesiredAgentImage                    string                         `json:"desiredAgentImage"`
+	ImportedConfig                       *ImportedConfig                `json:"importedConfig,omitempty" norman:"nocreate,noupdate"`
+	RancherKubernetesEngineConfig        *RancherKubernetesEngineConfig `json:"rancherKubernetesEngineConfig,omitempty"`
+	GenericEngineConfig                  *MapStringInterface            `json:"genericEngineConfig,omitempty"`
```
cjellick: add old types as map string interface. also, drop the fields from the schema. and add a controller that copies the data into generic field.

apis/management.cattle.io/v3/schema_types.go
nathan-jenan-rancher: When dynamic schemas are created or updated the schema created from the dynamic schema is merged with the existing schema.  The problem with this is that if a field is removed from the dynamic schema that field is not removed from the existing schema.  This makes it impossible to remove a dynamic field from a schema.

If dynamic field is set to true, then that field is removed from the existing schema before it is merged with the new schema.

The reason it has to exist, and isn't always just assumed to be true implicitly, is because the dynamic schema is actually marshaled to json, and then unmarshaled to the schema type. This only works because schema and dynamic schema are functionally identical despite being completely different types.  So while dynamic field will always be set to `True` on dynamic schema, it will be set to false on `Schema`.

ibuildthecloud:
LGTM

apis/management.cattle.io/v3/schema/schema.go
nathan-jenan-rancher: No, reverted this change.

apis/management.cattle.io/v3/schema/schema.go
nathan-jenan-rancher: Done.

client/management/v3/zz_generated_kontainer_driver.go
nathan-jenan-rancher: Marked as `noupdate`.
Added logic to formatter.

apis/management.cattle.io/v3/cluster_types.go
nathan-jenan-rancher: Done.

PR #419, opened 2018-05-07
Improve logging

aiwantaozi:
related issue: https://github.com/rancher/rancher/issues/13268
1. add backoff when deploy embedded failed
2. add watcher to check the embedded endpoint status
3. why roll back get endpoint from pod instead of endpoint because pod has more clear info why deploy failed, like 'Insufficient memory', and will show the same message as workload.


apis/management.cattle.io/v3/logging_types.go
```diff
@@ -63,8 +63,9 @@ type ProjectLoggingSpec struct {
 }
 
 type ClusterLoggingStatus struct {
-	Conditions  []LoggingCondition `json:"conditions,omitempty"`
-	AppliedSpec ClusterLoggingSpec `json:"appliedSpec,omitempty"`
+	Conditions  []LoggingCondition  `json:"conditions,omitempty"`
+	AppliedSpec ClusterLoggingSpec  `json:"appliedSpec,omitempty"`
+	FailedSpec  *ClusterLoggingSpec `json:"failedSpec,omitempty"`
```
alena1108: This field doesn't seem to be used in corresponding Rancher PR; can we remove it for now?

aiwantaozi: FailedSpec and AppliedSpec are used to prevent the infinite loop, the condition update will also trigger the sync function, once the AppliedSpec equal to the Spec, will not continue. The FailedSpec is for store the spec apply failed, if the same spec applies failed, will back off, and sync after the backoff time.

PR #421, opened 2018-05-08
Addon retries

moelsayed:
https://github.com/rancher/rke/issues/583
https://github.com/rancher/rke/pull/589


apis/management.cattle.io/v3/rke_types.go
```diff
@@ -35,6 +35,8 @@ type RancherKubernetesEngineConfig struct {
 	CloudProvider CloudProvider `yaml:"cloud_provider" json:"cloudProvider,omitempty"`
 	// kubernetes directory path
 	PrefixPath string `yaml:"prefix_path" json:"prefixPath,omitempty"`
+	// Number of status check retries for addon deployment jobs
+	AddonJobRetries int `yaml:"addon_job_retries" json:"addonJobRetries,omitempty"`
```
alena1108: what would be the default value? 0?

moelsayed: default is 5, set in rke

PR #425, opened 2018-05-10
Protected role templates

nathan-jenan-rancher:
This change adds a disabled field to role templates so admins can prevent them from being assigned to users.

Issue:
https://github.com/rancher/rancher/issues/13418


cjellick:
LGTM

PR #429, opened 2018-05-15
Two issues: delete noupdate on enabled field of AuthConfig + delete Admin field in AzureADConfig

carolynhu:
https://github.com/rancher/rancher/issues/12942


cjellick:
LGTM

PR #430, opened 2018-05-16
Remove dnsLabel for smtp host

zionwu:
Related issues: https://github.com/rancher/rancher/issues/13516

### DIFF
--- a/apis/management.cattle.io/v3/alerting_types.go
+++ b/apis/management.cattle.io/v3/alerting_types.go
@@ -133,7 +133,7 @@ type Notification struct {
 }
 
 type SMTPConfig struct {
-	Host             string `json:"host,omitempty" norman:"required,type=dnsLabel"`
+	Host             string `json:"host,omitempty" norman:"required"`
 	Port             int    `json:"port,omitempty" norman:"required,min=1,max=65535,default=587"`
 	Username         string `json:"username,omitempty"`
 	Password         string `json:"password,omitempty"`


vincent99:
The correct type is `hostname`; a hostname is a series of `dnsLabel`s separated by dots, a `dnsLabe is a single section and cannot have dot in it.

aiwantaozi:
LGTM

vincent99:
No, it doesn't look good.  The right type is`hostname`, a `string` is just going to have no validation at all.

alena1108:
@zionwu @vincent99 if we are to introduce `hostname` type, norman should be fixed to do the parsing/validation similar to dnsLabel: https://github.com/rancher/norman/blob/4cf9f645cd547793e52b9b2786faca9e9b1a4e64/parse/builder/builder.go#L263

zionwu:
@vincent99 @alena1108 
In the builder.go , we can find that `dnsLable` is using `validation.IsDNS1123Subdomain`, which is https://github.com/rancher/norman/blob/master/vendor/k8s.io/apimachinery/pkg/util/validation/validation.go#L134. 

I think the validation works fine for hostname. Maybe we should change the tag name from `dnsLabel` to `subdomain`. 

The issue is a UI issue. I will close this PR.

vincent99:
This is _not_ a UI issue.  

A hostname is a series of labels, separated by dots.  A label by definition cannot contain dots.  We called the type `dnsLabel` to disambiguate from generic "labels" on all resources.

These are not new terms, we agreed to them [6 months ago](https://github.com/rancher/ember-api-store/commit/9bf13030dcc5d552a0944dce1b7944a2e2eec024#diff-7bfc04b9085824990ffb1ee13b399590) and added support in the UI, which was eventually broken such that the UI was not showing any errors it found.

There are 3 issues, all in the API:

- The `type` of the field is wrong, `smtpHost` should be `type=hostname`
- The validation in the API for `dnsLabel` is wrong, you are validating that they are a hostname, not a label.
- There is apparently no support or validation at all for the correct type `hostname`.

zionwu:
@vincent99, ok, I will have a PR for norman framework and then one for types.

PR #441, opened 2018-05-18
Change from dnsLabel to hostname for smtp host

zionwu:
Relate issues:
rancher/rancher#13516

Related PR:
https://github.com/rancher/norman/pull/164


aiwantaozi:
LGTM, but the validation function for dnsLabel changed, need to let others check which type they want, dnsLabel or hostName.

zionwu:
Close as this is included in https://github.com/rancher/types/pull/442

PR #443, opened 2018-05-18
Revert azuread in API

carolynhu:


apis/management.cattle.io/v3/schema/schema.go
```diff
@@ -325,22 +325,7 @@ func authnTypes(schemas *types.Schemas) *types.Schemas {
 			}
 			schema.CollectionMethods = []string{}
 			schema.ResourceMethods = []string{http.MethodGet, http.MethodPut}
-		}).
-		MustImport(&Version, v3.ActiveDirectoryTestAndApplyInput{}).
```
deniseschannon: This line doesn't seem like it should be removed.

deniseschannon:
LGTM

PR #444, opened 2018-05-18
Make group member mapping attribute for AD required

mrajashree:
And set its default to `member`. This attribute will be used in fetching
user's groups, including nested groups


cjellick:
LGTM

PR #457, opened 2018-05-31
Make Cluster displayName a required field

prachidamle:
Schema change to mark clusterSpec.displayName required.

https://github.com/rancher/rancher/issues/13075


prachidamle:
@alena1108 review

PR #465, opened 2018-06-07
add desiredNodeUnschedulable to NodeSpec

kinarashah:
https://github.com/rancher/rancher/issues/13836
https://github.com/rancher/rancher/issues/13623

### DIFF
--- a/apis/management.cattle.io/v3/machine_types.go
+++ b/apis/management.cattle.io/v3/machine_types.go
@@ -177,15 +177,16 @@ type NodeSpec struct {
 	Worker           bool   `json:"worker" norman:"noupdate"`
 	NodeTemplateName string `json:"nodeTemplateName,omitempty" norman:"type=reference[nodeTemplate],noupdate"`
 
-	NodePoolName           string            `json:"nodePoolName" norman:"type=reference[nodePool],nocreate,noupdate"`
-	CustomConfig           *CustomConfig     `json:"customConfig"`
-	Imported               bool              `json:"imported"`
-	Description            string            `json:"description,omitempty"`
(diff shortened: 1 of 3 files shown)


apis/management.cattle.io/v3/machine_types.go
```diff
@@ -177,15 +177,16 @@ type NodeSpec struct {
...
+	NodePoolName             string            `json:"nodePoolName" norman:"type=reference[nodePool],nocreate,noupdate"`
+	CustomConfig             *CustomConfig     `json:"customConfig"`
+	Imported                 bool              `json:"imported"`
+	Description              string            `json:"description,omitempty"`
+	DisplayName              string            `json:"displayName"`
+	RequestedHostname        string            `json:"requestedHostname,omitempty" norman:"type=dnsLabel,nullable,noupdate,required"`
+	InternalNodeSpec         v1.NodeSpec       `json:"internalNodeSpec"`
+	DesiredNodeLabels        map[string]string `json:"desiredNodeLabels,omitempty"`
+	DesiredNodeAnnotations   map[string]string `json:"desiredNodeAnnotations,omitempty"`
+	DesiredNodeUnschedulable map[string]bool   `json:"desiredNodeUnschedulable,omitempty"`
```
alena1108: Does this field get updated via cordon/uncordon action? if so, we should mark it as nocreate/noupdate so api user doesn't set it directly. Also why is that a map? looks like it should be bool

kinarashah: - made `unschedulable` `nocreate,noupdate` 
- used string field to indicate `""` not being set

PR #467, opened 2018-06-08
add ssl

aiwantaozi:
add ssl support for logging target elasticsearch, splunk and kafka

### DIFF
--- a/apis/management.cattle.io/v3/logging_types.go
+++ b/apis/management.cattle.io/v3/logging_types.go
@@ -40,9 +40,9 @@ type ProjectLogging struct {
 type LoggingCommonSpec struct {
 	DisplayName string `json:"displayName,omitempty"`
 
-	OutputFlushInterval int               `json:"outputFlushInterval,omitempty" norman:"default=3"`
-	OutputTags          map[string]string `json:"outputTags,omitempty"`
-
+	OutputFlushInterval int                  `json:"outputFlushInterval,omitempty" norman:"default=3"`
+	OutputTags          map[string]string    `json:"outputTags,omitempty"`
+	DockerRootDir       string               `json:"dockerRootDir" norman:"default=/var/lib/docker/containers"`
 	ElasticsearchConfig *ElasticsearchConfig `json:"elasticsearchConfig,omitempty"`
 	SplunkConfig        *SplunkConfig        `json:"splunkConfig,omitempty"`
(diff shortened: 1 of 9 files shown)


apis/management.cattle.io/v3/logging_types.go
```diff
@@ -94,19 +94,27 @@ type LoggingCondition struct {
...
+	Endpoint      string `json:"endpoint,omitempty" norman:"required"`
+	IndexPrefix   string `json:"indexPrefix,omitempty" norman:"required"`
+	DateFormat    string `json:"dateFormat,omitempty" norman:"required,type=enum,options=YYYY-MM-DD|YYYY-MM|YYYY,default=YYYY-MM-DD"`
+	AuthUserName  string `json:"authUsername,omitempty"` //secret
+	AuthPassword  string `json:"authPassword,omitempty"` //secret
+	Certificate   string `json:"certificate"`
+	ClientCert    string `json:"clientCert"`
+	ClientKey     string `json:"clientKey"`
+	ClientKeyPass string `json:"clientKeyPass"`
+	SSLVerify     bool   `json:"sslVerify" norman:"default=false"`
```
alena1108: default is always false for the boolean, so there is no need to define the norman tag

apis/management.cattle.io/v3/logging_types.go
```diff
@@ -124,6 +132,9 @@ type KafkaConfig struct {
 	ZookeeperEndpoint string   `json:"zookeeperEndpoint,omitempty"`
```
alena1108: same here, no need to define the default

apis/management.cattle.io/v3/logging_types.go
```diff
@@ -124,6 +132,9 @@ type KafkaConfig struct {
...
 	BrokerEndpoints   []string `json:"brokerEndpoints,omitempty"`
 	Topic             string   `json:"topic,omitempty" norman:"required"`
+	Certificate       string   `json:"certificate"`
+	ClientCert        string   `json:"clientCert"`
+	ClientKey         string   `json:"clientKey"`
 }
```
alena1108: no SSLVerify for Zookeeper? Just wanted to make sure it's intended, as other configs have this flag set.

aiwantaozi: Fluentd Kafka plugin does not support sslVerify yet, need to send PR to them, so not in this time

PR #473, opened 2018-06-12
Gke private cluster

nathan-jenan-rancher:
Adding fields to support GKE private clusters
    
Issue:
https://github.com/rancher/rancher/issues/13980


cjellick:
close in favor of https://github.com/rancher/kontainer-engine/pull/106

PR #474, opened 2018-06-13
Correct Openstack configuration section name

sebastien-prudhomme:
See this bug in RKE rancher/rke#689


sebastien-prudhomme:
Finally find a way to do the "go generate"

PR #476, opened 2018-06-19
Eks options

nathan-jenan-rancher:
Adding region, instance type, and min and max node options to EKS config.

Issue:
https://github.com/rancher/rancher/issues/14064


StrongMonkey:
lgtm

PR #483, opened 2018-06-25
fix capability mapping

StrongMonkey:


StrongMonkey:
This fixes the issue that exists in master now:
1. Create a workload in UI. The workload will have an empty capability field.
2. Try to scale up the workload. This will update the workload and then remove the capability field, causing pods getting recreated.
3. Try to edit workload. This will update workload and then add capability field back, also recreate pods.

The fix would be:
Check if capAdd or capDrop are empty, and if so, remove them. This will always remove capability field if both capAdd and capDrop are empty.

Downside:
When launching a workload through kubectl, then edit through rancher UI, there is a possibility that pods get recreated. This is because we always set affinity and securityContext to nil if we found it empty. This is to address a different bug where you can't remove scheduling rules. Related to [here](https://github.com/rancher/rancher/issues/13527)

alena1108:
LGTM

PR #490, opened 2018-07-03
Add setting to enable/disable nested group memberships for AD

mrajashree:
The setting is true by default
https://github.com/rancher/rancher/issues/14284


prachidamle:
types PR is LGTM

PR #498, opened 2018-07-11
Node drain input

kinarashah:
https://github.com/rancher/rancher/issues/14265


apis/management.cattle.io/v3/machine_types.go
```diff
@@ -272,3 +274,19 @@ type PublicEndpoint struct {
 	// True when endpoint is exposed on every node
 	AllNodes bool `json:"allNodes" norman:"nocreate,noupdate"`
 }
+
+type NodeDrainInput struct {
+	// Drain node even if there are pods not managed by a ReplicationController, Job, or DaemonSet
```
alena1108: is it a k8s native param? can we just embed their native type to this struct?

PR #504, opened 2018-07-16
Add cluster monitoring options and system image

galal-hussein:


apis/management.cattle.io/v3/rke_types.go
```diff
@@ -39,6 +39,8 @@ type RancherKubernetesEngineConfig struct {
 	AddonJobTimeout int `yaml:"addon_job_timeout" json:"addonJobTimeout,omitempty" norman:"default=30"`
 	// Bastion/Jump Host configuration
 	BastionHost BastionHost `yaml:"bastion_host" json:"bastionHost,omitempty"`
+	// Disable Metrics Server addon
+	DisableMetricsServer bool `yaml:"disable_metrics_server" json:"disableMetricsServer,omitempty"`
```
alena1108: why bool? can we make it similar to ingress? so it has a provider and a list of options

galal-hussein: hmmm, i see, okay will fix

apis/management.cattle.io/v3/rke_types.go
```diff
@@ -552,3 +556,11 @@ type AzureCloudProvider struct {
 // AWSCloudProvider options
 type AWSCloudProvider struct {
 }
+
+// MetricsServerConfig
+type MetricsServerConfig struct {
+	// Disable metrics server
+	Disable bool `yaml:"disable" json:"disable,omitempty"`
```
alena1108: can we just use None value to disable it? again, similar to other addons

alena1108: and whatever provider is used for the current metric server as an alternative

galal-hussein: the problem is that metrics server doesn't have plugins like ingress, but it does contain options

alena1108:
LGTM

PR #506, opened 2018-07-17
add ssl support for syslog

aiwantaozi:


alena1108:
@aiwantaozi looks like we need generated changes

PR #510, opened 2018-07-19
Update rke tools 4

galal-hussein:


deniseschannon:
https://github.com/rancher/rancher/issues/14613

PR #519, opened 2018-07-25
fix duplicate transitioningMessage

carolynhu:
https://github.com/rancher/rancher/issues/14658
need to update types reference in rancher/rancher


alena1108:
LGTM

PR #525, opened 2018-07-31
Check nativeNode data is nil before executing mapping

ukinau:


StrongMonkey:
LGTM. @ukinau Is there an use case that you need to call /v3/nodes to create nodes?

ukinau:
Yes there is.
In our usecase, sometimes we want to delete specific node after created cluster, nodes.

But currently if we create cluster with nodeDriver like OpenStack on GUI, nodepool is created/used for the cluster and node will be indirectly created by nodepool controller. This behaviour itself is good to us but the problem is that  it's not easy to delete specific node in the case of node which is created by nodepool. If we want to delete specific node which is not last node in nodepool, we need to do 3 things to delete. 
1. delete target node
2. wait until new node is created  by nodepool
3. decrease the quantity (delete the node created in step 2)
    -> if we didn't wait, we will unintentionally delete other node.

This is not convenient for us, That's why we started to consider to use /v3/nodes API directly.

PR #536, opened 2018-08-07
Use secretName instead of secretId

kinarashah:
Using reference to id always returns `namespace:id`, but it doesn't
work while using project scoped secrets; thus using name similar
to secrets used in environment variables from source

https://github.com/rancher/rancher/issues/14740
https://github.com/rancher/rancher/issues/14368


ibuildthecloud:
LGTM

PR #539, opened 2018-08-16
add cancel action on node

kinarashah:
https://github.com/rancher/rancher/issues/14265


vincent99:
That's not the greatest name for an action... `cancel` what?

Also "cancelling" usually implies there are no side-effects (i.e. click edit on something, change a bunch of fields, click cancel.. nothing actually happened).  Since the drain has already started and pods are already deleted/moved around, all you can do is "stop" a drain from continuing.  (so `stopDrain`?).

PR #541, opened 2018-08-16
Update k8s version to v1.11.2-rancher1-1

galal-hussein:


apis/management.cattle.io/v3/k8s_defaults.go
```diff
@@ -19,6 +19,7 @@ var (
 		"v1.9.7-rancher2-2",
 		"v1.10.5-rancher1-2",
 		"v1.11.1-rancher1-1",
```
deniseschannon: we should remove this one. there should only ever be 1 version per minor version of k8s

deniseschannon:
https://github.com/rancherlabs/the-ranch/issues/154

deniseschannon:
@galal-hussein can we make this into 2 separate PRs?

PR #546, opened 2018-08-21
enable keycloak auth

carolynhu:
for my testing use
In the future if we want to enable keycloak, we can use this PR


cjellick:
LGTM

PR #555, opened 2018-08-30
Aks vnet options

nathan-jenan-rancher:
Adding options to the AKS config so that options can be passed to support
setting a custom vnet (without these options, a cluster with a custom vnet
will not be functional).

Issue:
rancher/rancher#15135


cjellick:
LGTM

PR #560, opened 2018-08-31
change norman type from dnsLabel to hostname for RequestedHostname 

carolynhu:
https://github.com/rancher/rancher/issues/14571


alena1108:
@carolynhu could you explain the reason why we do it, for the future reference.

carolynhu:
Here we extend the restriction for `RequestedHostName` field from `dnsLabel` to `hostname`. Since `dnsLabel` does not allow .(dot) to be set in `RequestedHostName`. If we change it to `hostname`, we make it possible to update `labels`, `custom name` for node with hostname contains . (dot), such as `ip address`. Then we only need to make sure we validate dnsLabel restriction on hostname when node is created.

alena1108:
LGTM

PR #561, opened 2018-09-01
set defaults for rke cluster

kinarashah:
https://github.com/rancher/rancher/issues/15256


alena1108:
@kinarashah we should make network/auth fields required

PR #566, opened 2018-09-06
Remove setting clusterIP type for dnsRecord from Schema

orangedeng:
Related issue https://github.com/rancher/rancher/issues/14650
This logic is moved to store for now.


apis/project.cattle.io/v3/schema/schema.go
```diff
@@ -584,7 +584,6 @@ func addServiceOrDNSRecord(dns bool) types.SchemasInitFunc {
 		schemas = schemas.
 			AddMapperForType(&Version, v1.ServiceSpec{},
 				&m.Move{From: "externalName", To: "hostname"},
```
alena1108: shall we also make clusterIP field non-updatable? there is norman `noupdate` tag

apis/project.cattle.io/v3/schema/schema.go
```diff
@@ -584,7 +584,6 @@ func addServiceOrDNSRecord(dns bool) types.SchemasInitFunc {
...
-				&ServiceSpecMapper{},
```
alena1108: is service spec mapper used anywhere else? If not, lets clean it up

apis/project.cattle.io/v3/schema/schema.go
orangedeng: It is non-updatable already in https://github.com/rancher/types/blob/master/apis/project.cattle.io/v3/schema/schema.go#L625

orangedeng:
Updated

PR #535, opened 2018-08-03
fix regular expression

StrongMonkey:

PR #567, opened 2018-09-06
fix regresssion for scheduling

StrongMonkey:
https://github.com/rancher/rancher/issues/15379.
Fix the regression introduced by https://github.com/rancher/types/pull/535. We accidently remove the whitespace between in and notin.


StrongMonkey:
https://github.com/rancher/rancher/issues/15379

PR #569, opened 2018-09-11
fix namespace_reference mapper on update handler

gitlawr:
Fix https://github.com/rancher/rancher/issues/15481
Namespace_reference mapper trims the namespace prefix of the reference fields in `/v3/project` resources. When pipeline resources are moved to project level, and a pipeline in project ns may refer to a sourcecodecredential in user ns, the namespace prefix needs to be preserved in this case. Changes [here](https://github.com/rancher/types/commit/36f2de76ce05e3396ca1d1c7f297ff17f1b0b34f#diff-5d5b04314a33e0c0a419b5723a91cadfR34) check and trim the ns when reference namespace is the same.
However, on updates `noupdate` fields are dropped, therefore no namespaceId is available for the mapper to check. So in this pr it does the trim when namespaceId is missing and `noupdate` is added to credential reference to bypass the trim.


mapper/namespace_reference.go
```diff
@@ -32,13 +32,10 @@ func (n *NamespaceReference) FromInternal(data map[string]interface{}) {
...
 func (n *NamespaceReference) ToInternal(data map[string]interface{}) error {
 	namespaceID, ok := data["namespaceId"]
-	if !ok {
-		return nil
-	}
 	for _, path := range n.fields {
 		convert.Transform(data, path, func(input interface{}) interface{} {
 			parts := strings.SplitN(convert.ToString(input), ":", 2)
-			if len(parts) == 2 && (parts[0] == namespaceID) {
+			if len(parts) == 2 && (!ok || parts[0] == namespaceID) {
```
alena1108: @gitlawr this change doesn't seem to cover the case when namespaceID is missing and len(parts) != 2 - previously namespaceId check was covering this situation. Should we be concerned?

alena1108:
@gitlawr my concern is why did we trip the namespace from the links to begin with, what is the reason for this mapper. Are there places in the code relying on the certain format? Then split brain situation - when some links are stored one way, and other in another way - can it potentially introduce issues?

gitlawr:
Take PVC as the example, In `https://github.com/rancher/types/blob/master/apis/project.cattle.io/v3/schema/schema.go#L729` we overwrite `claimName` to add reference type in the schema. In Rancher API, we use `ns:pvcname` as the id to refer to a pvc. But in K8S it's the name without the namespace prefix. So the mapper helps for the conversion by tripping the namespace.

gitlawr:
I'm not sure if there's other reasons for the mapper but I think you're right, there might be potential issues if we store the links in different ways.

gitlawr:
It would be safer to revert the changes on the mapper and find other solutions to obey existing rules.
Two solutions for the pipeline-credential reference case that I can think of now,
1. move sourceCodeCredential/sourceCodeRepo resource out of project api group(to management group)
2. remove norman type tag of the credential reference so as to keep the namespace from being tripped.

alena1108:
@gitlawr 

> move sourceCodeCredential/sourceCodeRepo resource out of project api group(to management group)

does the object that references `sourceCodeCredential/sourceCodeRepo` exist in management or user group?

> remove norman type tag of the credential reference so as to keep the namespace from being tripped.

given this path, how the correct credential would be determined? relying on the name?

gitlawr:
@alena1108 
1. the object that references sourceCodeCredential is in the project group
2. it remains to be the id of a credential,`ns:name`.

gitlawr:
I think it would be more flexible and loosely coupled if we can have remove-mapper utility in norman.
I've submit a pr in norman https://github.com/rancher/norman/pull/196.
Then we can revert changes on namespace_reference mapper and move out the mapper in the pipeline case, as updated in this pr.

alena1108:
@gitlawr based on conversation with @ibuildthecloud, he is actually fine with the original approach - not trimming the ns if it is different from the original one - and doesn't think it would break any of the existing stuff - my main original concern.

PR #575, opened 2018-09-18
Add AD auth provider methods

leodotcloud:
https://github.com/rancher/rancher/issues/15693
Problem: Using the go client for automation doesn't provide for the methods to use AD
Fix: Add the generated methods


cjellick:
@leodotcloud I see the problem you are hitting with the generated code, but unfortunately I cant merge this as-is.
The API was intentionally designed to expose the methods in a very controlled fashion. It controls the login flow. The change you made would impact that

prachidamle:
@leodotcloud question out of curiosity: I see only change to the schema you did is adding the resourceMethod "http.MethodGet" 
and this generated all the methods you needed on that resource?

prachidamle:
ha, now I remember after reading @cjellick's comment. Yeah this was designed to not expose all APIs on the providers.

cjellick:
Closing. As discussed, this changes the API in an undesirable way.

PR #578, opened 2018-09-24
Cluster capabilities

mrajashree:
https://github.com/rancher/rancher/issues/12422


apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -283,3 +287,23 @@ type ImportClusterYamlInput struct {
 type ImportYamlOutput struct {
 	Message string `json:"message,omitempty"`
 }
+
+type Capabilities struct {
+	L4LoadBalancer           L4LoadBalancer    `json:"l4loadBalancer,omitempty"`
+	IngressController        IngressController `json:"ingressController,omitempty"`
```
alena1108: it should be an array of ingress controllers as there can be many

PR #580, opened 2018-09-27
Fix typo in weave_node rke_types

mickael-alibert:
There's an 'e' missing from weave_node name that impacts the possibility to define weave-node system-image version in RKE


galal-hussein:
@mickael-alibert can you run `go generate` on the PR

PR #591, opened 2018-10-10
Adding public ip option for worker nodes

nathan-jenan-rancher:
Adding option to associate worker node public ip to the eks config to support
private clusters in EKS.

Issue:
rancher/rancher#15419


StrongMonkey:
LGTM

PR #593, opened 2018-10-15
Eks Session token for Assume Role

nathan-jenan-rancher:
This change adds a session token field to the eks config to support
provisioning clusters with credentials generated via GetSessionToken or
AssumeRole calls.

Issue:
rancher/rancher#16005


StrongMonkey:
LGTM

PR #596, opened 2018-10-18
WIP: Add global DNS type

prachidamle:
Initial GlobalDNS type


apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -0,0 +1,15 @@
...
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+type GlobalDNS struct {
+	metav1.TypeMeta `json:",inline"`
+	// Standard object’s metadata. More info:
+	// https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#metadata
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	DNSName      string `json:"dnsName,omitempty" norman:"required"`
```
alena1108: do you think `name` can play the role of `fqdn`? So we can omit the DNSName all along. RootDomain is still relevant

prachidamle: yes makes sense

PR #598, opened 2018-10-23
Multi cluster app CRD

mrajashree:


apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   MultiClusterAppSpec   `json:"spec,omitempty"`
+	Status MultiClusterAppStatus `json:"status,omitempty"`
+}
+
+type MultiClusterAppSpec struct {
+	HelmChartReference HelmChartReference `json:"helmChartReference,omitempty"`
```
alena1108: we should remove `Reference` from names

apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+}
+
+type MultiClusterAppStatus struct {
+	HealthState string `json:"healthState,omitempty" norman:"required"`
+}
+
+type HelmChartReference struct {
+	ChartRepositoryURL string `json:"chartRepositoryUrl,omitempty"`
+	ChartReference     string `json:"chartReference,omitempty"`
+	ChartVersion       string `json:"chartVersion,omitempty"`
```
alena1108: Lets remove Chart from names; as HelmChart is already in a super type name

apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+	ReleaseName        string `json:"releaseName,omitempty"`
+	ReleaseNamespace   string `json:"releaseNamespace,omitempty"`
```
alena1108: Namespace should be a property of the target, not the template or helmchart

apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+}
+
+type TemplateReference struct {
+	TemplateName    string `json:"templateName,omitempty"`
```
alena1108: should be just name and version. Both should have norman reference tag on them

apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+	TemplateVersion string `json:"templateVersion,omitempty"`
+}
+
+type Target struct {
+	metav1.TypeMeta   `json:",inline"`
```
alena1108: why do we have type meta on the target? thought we do it only for CRDs?

apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   TargetSpec   `json:"spec,omitempty" norman:"required"`
+	Status TargetStatus `json:"status,omitempty" norman:"required"`
+}
+
+type TargetSpec struct {
+	ClusterName   string            `json:"clusterName,omitempty"`
```
alena1108: need a norman annotation as this is a reference type

apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+	Answers       map[string]string `json:"answers,omitempty"`
+}
+
+type TargetStatus struct {
+	ChartReleaseName string `json:"chartReleaseName,omitempty" norman:"required"`
+	HealthState      string `json:"healthState,omitempty" norman:"required"`
+}
+
+type ClusterConfig struct {
+	Namespace                string `json:"namespace,omitempty"`
```
alena1108: why do we have a namespace on a cluster config?

apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   TargetSpec   `json:"spec,omitempty" norman:"required"`
+	Status TargetStatus `json:"status,omitempty" norman:"required"`
+}
+
+type TargetSpec struct {
+	ClusterName   string            `json:"clusterName,omitempty"`
+	ClusterConfig ClusterConfig     `json:"clusterConfig,omitempty"`
+	Answers       map[string]string `json:"answers,omitempty"`
```
alena1108: Lets add answers on a global level as well. So user can define a super set on a global level, and override some of it on per cluster basis

apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+}
+
+type TargetStatus struct {
+	ChartReleaseName string `json:"chartReleaseName,omitempty" norman:"required"`
```
alena1108: would this parameter be common for a template based apps, and helmchart based ones? Looks like for template based ones, it can be an app id

apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+	HealthState      string `json:"healthState,omitempty" norman:"required"`
+}
+
+type ClusterConfig struct {
+	Namespace                string `json:"namespace,omitempty"`
+	Server                   string `json:"server,omitempty" norman:"required"`
+	CertificateAuthorityPath string `json:"certificateAuthorityPath,omitempty"`
```
alena1108: is there a need for so many parameters? does rancher resource have as many fields?

apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+
+	Spec   TargetSpec   `json:"spec,omitempty" norman:"required"`
+	Status TargetStatus `json:"status,omitempty" norman:"required"`
+}
+
+type TargetSpec struct {
+	ClusterName      string            `json:"clusterName,omitempty" norman:"type=reference[cluster]"`
+	ClusterConfig    ClusterConfig     `json:"clusterConfig,omitempty"`
+	Answers          map[string]string `json:"answers,omitempty"`
+	ReleaseNamespace string            `json:"releaseNamespace,omitempty"`
```
alena1108: should be just the namespace. And of the reference type

alena1108: Should also have a project id; also of a reference type

apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   MultiClusterAppSpec   `json:"spec,omitempty"`
+	Status MultiClusterAppStatus `json:"status,omitempty"`
+}
+
+type MultiClusterAppSpec struct {
+	ReleaseName string             `json:"releaseName,omitempty"`
+	Answers     map[string]string  `json:"answers,omitempty"`
+	HelmChart   HelmChartReference `json:"helmChart,omitempty"`
```
alena1108: `HelmChartReference` and `TemplateReference` should be off the pointer type, as only one or another can be present.

apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+	Template    TemplateReference  `json:"template,omitempty"`
+	Targets     []Target           `json:"targets,omitempty" norman:"required"`
+}
+
+type MultiClusterAppStatus struct {
+	HealthState string `json:"healthState,omitempty" norman:"required"`
```
alena1108: required should not be set on the status fields

apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+}
+
+type HelmChartReference struct {
+	RepositoryURL string `json:"chartRepositoryUrl,omitempty"`
```
alena1108: need to fix json tags to be the same as fields names

apis/management.cattle.io/v3/multi_cluster_app_types.go
```diff
@@ -13,3 +16,67 @@ type GlobalDNS struct {
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   TargetSpec   `json:"spec,omitempty" norman:"required"`
+	Status TargetStatus `json:"status,omitempty" norman:"required"`
+}
+
+type TargetSpec struct {
+	ClusterName      string            `json:"clusterName,omitempty" norman:"type=reference[cluster]"`
+	ClusterConfig    ClusterConfig     `json:"clusterConfig,omitempty"`
```
alena1108: ClusterConfig should be of a pointer type as it won't be required when the tool is deployed in rancher

PR #601, opened 2018-10-23
Pipeline notification

gitlawr:
Address issue https://github.com/rancher/rancher/issues/15665


apis/project.cattle.io/v3/pipeline_types.go
```diff
@@ -155,8 +156,20 @@ type PipelineSpec struct {
...
+	Notification *PipelineNotification `json:"notification,omitempty" yaml:"notification,omitempty"`
+}
+
+type PipelineNotification struct {
+	Recipients []Recipient   `json:"recipients,omitempty" yaml:"recipients,omitempty"`
+	Message    string        `json:"message,omitempty" yaml:"message,omitempty"`
+	Condition  stringorslice `json:"condition,omitempty" yaml:"condition,omitempty"`
+}
+
+type Recipient struct {
```
guangbochen: should this `Recipient` need to be align with alert [Recipient](https://github.com/rancher/types/blob/b2cf716a27cd736fb6f238ee5a134547ac003ef7/apis/management.cattle.io/v3/alerting_types.go#L62) as they are literally same types of receivers?

gitlawr: The reason I wrote in this way is that it will be tedious for users to write the notifier reference Id(namespace:n-abc)+notifier type in their pipelinefiles. So this is a simplified version that users put either notifier display name or notifier Id(n-abc), we will do the search and find the match no matter what type the notifier is.

PR #619, opened 2018-11-09
Add multiclusterapp crd 

mrajashree:


alena1108:
We need to add multiClusterApp ref field to the app object

PR #620, opened 2018-11-09
[WIP] Add backup types

moelsayed:


apis/management.cattle.io/v3/backup_types.go
```diff
@@ -0,0 +1,98 @@
...
+	// Access key ID
+	AccessKeyID string `yaml:"access_key_id" json:"accessKeyId,omitempty"`
+	// Secret access key
+	SecretAccesssKey string `yaml:"secret_access_key" json:"secretAccessKey,omitempty"`
+	// name of the bucket to use for backup
+	BucketName string `yaml:"bucket_name" json:"bucketName,omitempty"`
+	// AWS Region, AWS spcific
+	Region string `yaml:"region" json:"region,omitempty"`
+	// Endpoint is used if this is not an AWS API
+	Endpoint string `yaml:"endpoint" json:"endpoint"`
```
guangbochen: @moelsayed do u think we can add bool type value `enabled` to the S3BackupBackend struct, basically will need it for the rke input types validation of the backupBackend, otherwise, it will be a little bit tricky to define when those inputs should be checked, thanks.

PR #622, opened 2018-11-15
Add rotate certificate input type

galal-hussein:


apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -302,3 +303,8 @@ type IngressCapabilities struct {
 	IngressProvider      string `json:"ingressProvider,omitempty"`
 	CustomDefaultBackend bool   `json:"customDefaultBackend,omitempty"`
 }
+
+type RotateCertificateInput struct {
+	RotateCACertificates bool     `json:"rotateCACertificates,omitempty" norman:"default=false"`
```
alena1108: I don't think we need Rotate* prepending the fields; should be just CACertificates, and Components
default=false is not needed. All bools are false by default

apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -302,3 +303,8 @@ type IngressCapabilities struct {
...
+	RotateComponents     []string `json:"rotateComponents,omitempty"`
```
alena1108: Would empty list mean all components/services?

apis/management.cattle.io/v3/cluster_types.go
galal-hussein: sure i will do this change

apis/management.cattle.io/v3/cluster_types.go
galal-hussein: yeah that's what it means

alena1108: we need to list available values - etcd, control, etc - using `norman:"type=array[enum],options=....` annotation

apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -74,6 +74,7 @@ type ClusterSpec struct {
 	DefaultClusterRoleForProjectMembers  string                               `json:"defaultClusterRoleForProjectMembers,omitempty" norman:"type=reference[roleTemplate]"`
 	DockerRootDir                        string                               `json:"dockerRootDir,omitempty" norman:"default=/var/lib/docker"`
 	EnableNetworkPolicy                  *bool                                `json:"enableNetworkPolicy" norman:"default=false"`
+	RotateCertificates                   *RotateCertificateInput              `json:"rotateCertificates,omitempty"`
```
alena1108: should rotateCertificates be a part of rkeConfig? as it's RKE clusters feature only

PR #633, opened 2018-11-26
Backup types

moelsayed:


apis/management.cattle.io/v3/backup_types.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"github.com/rancher/norman/condition"
+	"github.com/rancher/norman/types"
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	BackupConditionCreated    condition.Cond = "Created"
+	BackupConditionCompeleted condition.Cond = "Compeleted"
```
alena1108: typo in completed

apis/management.cattle.io/v3/backup_types.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+type BackupBackend struct {
+	// backend name
+	Name string `yaml:"name" json:"name,omitempty" norman:"default=s3"`
```
alena1108: Lets think if we really need a name...

apis/management.cattle.io/v3/backup_types.go
```diff
@@ -0,0 +1,72 @@
...
+	// s3 backend
+	S3BackupBackend *S3BackupBackend `yaml:",omitempty" json:"s3BackupBackend"`
+}
+
+type S3BackupBackend struct {
+	// Access key ID
+	AccessKeyID string `yaml:"access_key_id" json:"accessKeyId,omitempty"`
+	// Secret access key
+	SecretAccesssKey string `yaml:"secret_access_key" json:"secretAccessKey,omitempty"`
```
alena1108: should add norman annotation `password`

apis/management.cattle.io/v3/backup_types.go
```diff
@@ -0,0 +1,72 @@
...
+	// name of the bucket to use for backup
+	BucketName string `yaml:"bucket_name" json:"bucketName,omitempty"`
+	// AWS Region, AWS spcific
+	Region string `yaml:"region" json:"region,omitempty"`
+	// Endpoint is used if this is not an AWS API
+	Endpoint string `yaml:"endpoint" json:"endpoint"`
+}
```
alena1108: if you want custom name for the backups, you can add prefix as a field here

apis/management.cattle.io/v3/backup_types.go
```diff
@@ -0,0 +1,63 @@
...
+type BackupTarget struct {
+	// s3 target
+	S3BackupTarget *S3BackupTarget `yaml:",omitempty" json:"s3BackupTarget"`
+}
+
+type S3BackupTarget struct {
+	// Access key ID
+	AccessKeyID string `yaml:"access_key_id" json:"accessKeyId,omitempty"`
+	// Secret access key
+	SecretAccesssKey string `yaml:"secret_access_key" json:"secretAccessKey,omitempty" norman:"required,type=password" `
```
vincent99: JSON/YAML: We call these `accessKey` and `secretKey` (`access_key` / `secret_key`) in other places of the API (EC2 and EKS)

apis/management.cattle.io/v3/backup_types.go
```diff
@@ -0,0 +1,63 @@
...
+	Endpoint string `yaml:"endpoint" json:"endpoint"`
+}
+type EtcdBackup struct {
+	types.Namespaced
+
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// actual file name on the target
+	FileName string `yaml:"file_name" json:"fileName,omitempty"`
```
vincent99: JSON/YAML: `filename`

apis/management.cattle.io/v3/rke_types.go
```diff
@@ -183,6 +183,8 @@ type ETCDService struct {
 	Retention string `yaml:"retention" json:"retention,omitempty" norman:"default=72h"`
 	// Etcd snapshot Creation period
 	Creation string `yaml:"creation" json:"creation,omitempty" norman:"default=12h"`
```
vincent99: It thought we were going to fix this and Retention to be an integer number of minutes?  Is that a future release?

Don't like string times with complicated parsing and validation.

moelsayed: We will move this to a future iteration. For now, UI can set this to a number of hours.

apis/management.cattle.io/v3/backup_types.go
```diff
@@ -0,0 +1,64 @@
...
+type BackupTarget struct {
+	// s3 target
+	S3BackupTarget *S3BackupTarget `yaml:",omitempty" json:"s3BackupTarget"`
+}
+
+type S3BackupTarget struct {
+	// Access key ID
+	AccessKey string `yaml:"access_key" json:"accessKey,omitempty"`
+	// Secret access key
+	SecretKey string `yaml:"secret_key" json:"secretKey,omitempty" norman:"required,type=password" `
```
cjellick: @kinarashah will this be stored as a k8s secret out of the gate?

apis/management.cattle.io/v3/schema/schema.go
```diff
@@ -178,6 +180,10 @@ func clusterTypes(schemas *types.Schemas) *types.Schemas {
 				Input: "monitoringInput",
 			}
 			schema.ResourceActions["disableMonitoring"] = types.Action{}
+			schema.ResourceActions["saveSnapshot"] = types.Action{}
```
alena1108: Lets change it to `backupEtcd`

apis/management.cattle.io/v3/rke_types.go
```diff
@@ -185,6 +187,8 @@ type ETCDService struct {
...
+	// Backup backend for etcd snapshots, used by rke only
+	BackupTarget *BackupTarget `yaml:"backup_target" json:"backupTarget,omitempty"`
```
alena1108: Lets rename to the `BackupConfig`

apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -231,6 +231,11 @@ type IngressCapabilities struct {
 type MonitoringInput struct {
 	Answers map[string]string `json:"answers,omitempty"`
 }
+
+type RestoreSnapshotInput struct {
+	SnapshotName string        `json:"snapshotName,omitempty" norman:"type=reference[etcdBackup]"`
```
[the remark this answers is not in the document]
alena1108: lets rename it to etcdBackupName

apis/management.cattle.io/v3/rke_types.go
alena1108: update: we are introducing retention and intervalHours parameters to the BackupConfig as per conversation with @vincent99 : https://github.com/rancher/types/pull/633/commits/487bca05840ccf16008babd4ade145541e6ce803#diff-623d9a24d8befd392620d1315c53f7d2R18

apis/management.cattle.io/v3/backup_types.go
kinarashah: Yeah, have a pr coming in with some minor changes, with that it should work out of the gate

apis/management.cattle.io/v3/backup_types.go
```diff
@@ -0,0 +1,68 @@
...
+	// name of the bucket to use for backup
+	BucketName string `yaml:"bucket_name" json:"bucketName,omitempty"`
+	// AWS Region, AWS spcific
+	Region string `yaml:"region" json:"region,omitempty"`
+	// Endpoint is used if this is not an AWS API
+	Endpoint string `yaml:"endpoint" json:"endpoint"`
+}
+type EtcdBackup struct {
+	types.Namespaced
```
alena1108: we need to add a clusterid as a field

alena1108: This field is so the UI can filter out the etcd backups on per cluster. And on the rancher api side, we should error out if we try to recover to the backups from other clusters.

apis/management.cattle.io/v3/rke_types.go
alena1108: @guangbochen We wanted to introduce a new format for these 2 parameters (hours for creation, number of snapshots for retention). But in order to not break api compatibility with the previous versions, we couldn't change the existing params, so we've introduced new ones in backupConfig, and those params would be applied to S3 backups (for now) only. 

> If both ETCDService and BackupConfig has Retention set with 12h and 6(num of backups), which one it should follow when removing the local backups

Old parameters should apply to the local backups only; new params from backupConfig should apply to the backups on S3.

Let me know if everything is clear @guangbochen 

 @moelsayed please confirm

guangbochen: sure thanks @alena1108, I will add the s3 `backupConfig` to the rke-tools.

alena1108: @guangbochen we don't need to add it to the rke tools; we just have to change rke side code to parse the correct retention and creation parameters.

alena1108:
LGTM

PR #629, opened 2018-11-20
Import prometheus-operator CRDs into Rancher schema

orangedeng:
Related issue: https://github.com/rancher/rancher/issues/14230

PR #636, opened 2018-11-28
Rancher Monitoring Support

orangedeng:
Related Issue: https://github.com/rancher/rancher/issues/14230
Related PR: https://github.com/rancher/types/pull/629

This is a PR for rancher monitoring support. Summaries are follow:
1. Support prometheus crds in Rancher API.
2. Adding new kind of Alert feature that supports grouping alert rules.
3. Enable and disable cluster and project monitoring.
4. Support monitoring metrics graph.
5. Support prometheus to collect workload metrics with operator


apis/management.cattle.io/v3/authz_types.go
```diff
@@ -52,6 +54,7 @@ type ProjectSpec struct {
 	ClusterName                   string                  `json:"clusterName,omitempty" norman:"required,type=reference[cluster]"`
 	ResourceQuota                 *ProjectResourceQuota   `json:"resourceQuota,omitempty"`
 	NamespaceDefaultResourceQuota *NamespaceResourceQuota `json:"namespaceDefaultResourceQuota,omitempty"`
+	EnableProjectMonitoring       *bool                   `json:"enableProjectMonitoring" norman:"default=false"`
```
ibuildthecloud: why is this a pointer?

apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -74,6 +76,7 @@ type ClusterSpec struct {
 	DefaultClusterRoleForProjectMembers  string                               `json:"defaultClusterRoleForProjectMembers,omitempty" norman:"type=reference[roleTemplate]"`
 	DockerRootDir                        string                               `json:"dockerRootDir,omitempty" norman:"default=/var/lib/docker"`
 	EnableNetworkPolicy                  *bool                                `json:"enableNetworkPolicy" norman:"default=false"`
+	EnableClusterMonitoring              *bool                                `json:"enableClusterMonitoring" norman:"default=false"`
```
ibuildthecloud: why is this a pointer also

apis/management.cattle.io/v3/alerting_types.go
```diff
@@ -99,6 +99,155 @@ type AlertStatus struct {
 	AlertState string `json:"alertState,omitempty" norman:"options=active|inactive|alerting|muted,default=active"`
 }
 
+type ClusterAlertGroup struct {
```
cjellick: Where are all these types used? Are there rancher controllers that act on these?

apis/management.cattle.io/v3/authz_types.go
thxCode: similar to `Cluster` struct.

apis/management.cattle.io/v3/cluster_types.go
thxCode: similar to `EnableNetworkPolicy`, keep them in the same style.

apis/management.cattle.io/v3/authz_types.go
cjellick: We only do pointer bools when we _need_ to. That generally is to support backwards compatibility and upgrade scenarios. If it doesn't have a functional need to be a pointer, it shouldn't be.

cjellick:
Regarding the alert group changes, are you introducing any breaking API changes like removing or renaming a field? As best I can tell from reviewing, you are not, but I wanted to ask the question explicitly for your consideration.

apis/management.cattle.io/v3/schema/schema.go
```diff
@@ -501,12 +512,8 @@ func alertTypes(schema *types.Schemas) *types.Schemas {
 	return schema.
 		AddMapperForType(&Version, v3.Notifier{},
 			m.DisplayName{}).
-		AddMapperForType(&Version, v3.ClusterAlert{},
-			&m.Embed{Field: "status"},
-			m.DisplayName{}).
-		AddMapperForType(&Version, v3.ProjectAlert{},
-			&m.Embed{Field: "status"},
-			m.DisplayName{}).
+		MustImport(&Version, v3.ClusterAlert{}).
```
cjellick: why did you drop the embedding of status and the displayName mapper here?

apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -302,3 +306,7 @@ type IngressCapabilities struct {
 	IngressProvider      string `json:"ingressProvider,omitempty"`
 	CustomDefaultBackend bool   `json:"customDefaultBackend,omitempty"`
 }
+
+type MonitoringInput struct {
+	Answers map[string]string `json:"answers,omitempty"`
```
cjellick: What is the UX going to be around this? How are we going to present the questions and answers?

apis/management.cattle.io/v3/authz_types.go
thxCode: @cjellick , Got.

apis/management.cattle.io/v3/cluster_types.go
thxCode: This struct uses to overwrite the Answers of `cluster-monitoring` App and `project-monitoring` App, and we only allow users to overwrite some important fields, just like Prometheus retention, Storage Class, Storage Size and Node Exporter Port.

cjellick:
LGTM

PR #642, opened 2018-12-03
Add changes for auth refresh

maxkorp:
This allows tokens to be disabled along with tracking info for doing auth refreshes.

### DIFF
--- a/apis/management.cattle.io/v3/authn_types.go
+++ b/apis/management.cattle.io/v3/authn_types.go
@@ -25,6 +25,7 @@ type Token struct {
 	Expired         bool              `json:"expired"`
 	ExpiresAt       string            `json:"expiresAt"`
 	Current         bool              `json:"current"`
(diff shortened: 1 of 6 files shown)


apis/management.cattle.io/v3/authn_types.go
```diff
@@ -25,6 +25,7 @@ type Token struct {
 	Expired         bool              `json:"expired"`
 	ExpiresAt       string            `json:"expiresAt"`
 	Current         bool              `json:"current"`
+	Enabled         *bool             `json:"enabled,omitempty" norman:"default=true"`
```
cjellick: @maxkorp - make sure we have a separate issue to track the fact that you can enable/disable tokens. this is a small enhancment in and of itself and should be tested explicitly. also, open a UI issue so that they can add support to disabling tokens to the UI

cjellick:
LGTM

PR #647, opened 2018-12-07
Not omit threshold field

aiwantaozi:
Problem:
1. default value for float return zero string
2. we need threshold value to build alert expression, 0 is useful value

Solution:
1. fix float type default value in norman
2. not omit threshold value

Norman:
https://github.com/rancher/norman/pull/237

### DIFF
--- a/apis/management.cattle.io/v3/alerting_types.go
+++ b/apis/management.cattle.io/v3/alerting_types.go
@@ -200,12 +200,10 @@ type CommonRuleField struct {
 
 type MetricRule struct {
 	Expression     string  `json:"expression,omitempty" norman:"required"`
-	LegendFormat   string  `json:"legendFormat,omitempty"`
-	Step           int64   `json:"step,omitempty"`
 	Description    string  `json:"description,omitempty"`
-	Duration       string  `json:"duration,omitempty"`
+	Duration       string  `json:"duration,omitempty" norman:"required"`
 	Comparison     string  `json:"comparison,omitempty" norman:"type=enum,options=equal|not-equal|greater-than|less-than|greater-or-equal|less-or-equal,default=equal"`
-	ThresholdValue float64 `json:"thresholdValue,omitempty" norman:"type=float"`
+	ThresholdValue float64 `json:"thresholdValue,omitempty" norman:"required,type=float"`
 }
 
 type TimingField struct {
--- a/client/management/v3/zz_generated_metric_rule.go
+++ b/client/management/v3/zz_generated_metric_rule.go
@@ -6,8 +6,6 @@ const (
 	MetricRuleFieldDescription    = "description"
 	MetricRuleFieldDuration       = "duration"
 	MetricRuleFieldExpression     = "expression"
-	MetricRuleFieldLegendFormat   = "legendFormat"
-	MetricRuleFieldStep           = "step"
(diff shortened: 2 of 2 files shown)


apis/management.cattle.io/v3/alerting_types.go
```diff
@@ -201,11 +201,10 @@ type CommonRuleField struct {
 type MetricRule struct {
 	Expression     string  `json:"expression,omitempty" norman:"required"`
 	LegendFormat   string  `json:"legendFormat,omitempty"`
-	Step           int64   `json:"step,omitempty"`
 	Description    string  `json:"description,omitempty"`
 	Duration       string  `json:"duration,omitempty"`
 	Comparison     string  `json:"comparison,omitempty" norman:"type=enum,options=equal|not-equal|greater-than|less-than|greater-or-equal|less-or-equal,default=equal"`
-	ThresholdValue float64 `json:"thresholdValue,omitempty" norman:"type=float"`
+	ThresholdValue float64 `json:"thresholdValue" norman:"type=float"`
```
orangedeng: Should keep `omitempty`

orangedeng: I think it is fine it do not need to has the omit flag in json. LGTM.

orangedeng:
LGTM

apis/management.cattle.io/v3/alerting_types.go
```diff
@@ -200,12 +200,10 @@ type CommonRuleField struct {
...
 type MetricRule struct {
 	Expression     string  `json:"expression,omitempty" norman:"required"`
-	LegendFormat   string  `json:"legendFormat,omitempty"`
-	Step           int64   `json:"step,omitempty"`
 	Description    string  `json:"description,omitempty"`
-	Duration       string  `json:"duration,omitempty"`
-	Comparison     string  `json:"comparison,omitempty" norman:"type=enum,options=equal|not-equal|greater-than|less-than|greater-or-equal|less-or-equal,default=equal"`
-	ThresholdValue float64 `json:"thresholdValue,omitempty" norman:"type=float"`
+	Duration       string  `json:"duration,omitempty" norman:"required"`
+	Comparison     string  `json:"comparison,omitempty" norman:"required,type=enum,options=equal|not-equal|greater-than|less-than|greater-or-equal|less-or-equal,default=equal"`
```
alena1108: is `required` flag needed when the `default=equal` is set?

apis/management.cattle.io/v3/alerting_types.go
```diff
@@ -200,12 +200,10 @@ type CommonRuleField struct {
...
+	ThresholdValue float64 `json:"thresholdValue,omitempty" norman:"required,type=float"`
```
alena1108: do you think we can specify the default value for this parameter?

aiwantaozi: This value depends on the expression, like is the express etcd leader not equal 1, the threshold is 1, the memory usage close to 80%, the thresholds is 0.8, so it hard to define a default value for it.

apis/management.cattle.io/v3/alerting_types.go
aiwantaozi: updated

PR #648, opened 2018-12-07
Basic options and types for CIS best practice

jgreat:
* Add RKE types option for passing back "--dind" install

Update K8sVersionServiceOptions
* kubelet cadvisor-port moved from "default list" to version specific list.
  * Remove from 1.12 (no longer a valid option)
  * Add default "0" to 1.9 to 1.11
  * Note not valid in 1.8
* Add 1.8 block since 1.8 is still buildable with RKE

### DIFF
--- a/apis/management.cattle.io/v3/k8s_defaults.go
+++ b/apis/management.cattle.io/v3/k8s_defaults.go
@@ -35,7 +35,6 @@ var (
 			},
 			Kubelet: map[string]string{
 				"tls-cipher-suites": "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305",
-				"cadvisor-port":     "",
 			},
 		},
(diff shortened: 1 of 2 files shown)


apis/management.cattle.io/v3/rke_types.go
```diff
@@ -23,6 +23,8 @@ type RancherKubernetesEngineConfig struct {
 	Authorization AuthzConfig `yaml:"authorization" json:"authorization,omitempty"`
 	// Enable/disable strict docker version checking
 	IgnoreDockerVersion bool `yaml:"ignore_docker_version" json:"ignoreDockerVersion" norman:"default=true"`
+	// Is this a DinD install?
+	DinD bool `yaml:"dind" json:"dind"`
```
galal-hussein: its probably a better to add omitempty to the yaml annotation to avoid confusion when the user run `rke config`

apis/management.cattle.io/v3/k8s_defaults.go
```diff
@@ -55,13 +55,23 @@ var (
...
 		"v1.9": {
 			KubeAPI: map[string]string{
 				"endpoint-reconciler-type": "lease",
 				"admission-control":        "ServiceAccount,NamespaceLifecycle,LimitRanger,PersistentVolumeLabel,DefaultStorageClass,ResourceQuota,DefaultTolerationSeconds",
 			},
+			Kubelet: map[string]string{
+				"cadvisor-port": "0",
+			},
+		},
+		"v1.8": {
```
alena1108: we no longer support 1.8, so can remove this section

alena1108:
This pr will be consolidated in https://github.com/rancher/types/pull/650

PR #649, opened 2018-12-10
CloudCredential type 

kinarashah:
add new Cloud credential type that will proxy to management secrets and node templates would have reference to cloud credential


apis/management.cattle.io/v3/machine_types.go
```diff
@@ -43,6 +43,7 @@ type NodeTemplateSpec struct {
 	DisplayName      string `json:"displayName"`
 	Description      string `json:"description"`
 	Driver           string `json:"driver" norman:"nocreate,noupdate"`
+	CredentialName   string `json:"credentialName" norman:"type=reference[cloudCredential]"`
```
alena1108: should that be CloudCredentialName?

PR #650, opened 2018-12-10
WIP - CIS Benchmark Options - Admissions plugin AlwaysPullImages

jgreat:
Add option to kube-apiserver to enable AlwaysPullImages admissions plugin.

>Setting admission control policy to AlwaysPullImages forces every new pod to pull the
required images every time. In a multi-tenant cluster users can be assured that their
private images can only be used by those who have the credentials to pull them.


apis/management.cattle.io/v3/k8s_defaults.go
```diff
@@ -55,13 +55,23 @@ var (
...
 		"v1.9": {
 			KubeAPI: map[string]string{
 				"endpoint-reconciler-type": "lease",
 				"admission-control":        "ServiceAccount,NamespaceLifecycle,LimitRanger,PersistentVolumeLabel,DefaultStorageClass,ResourceQuota,DefaultTolerationSeconds",
 			},
+			Kubelet: map[string]string{
+				"cadvisor-port": "0",
+			},
+		},
+		"v1.8": {
```
alena1108: we no longer support k8s 1.8, so that can be dropped

.gitignore
```diff
@@ -7,3 +7,4 @@
 /trash.lock
 /types
 *trash.lock
+/.vscode
```
alena1108: Nit: new line

apis/management.cattle.io/v3/rke_types.go
```diff
@@ -194,6 +196,8 @@ type KubeAPIService struct {
 	ServiceNodePortRange string `yaml:"service_node_port_range" json:"serviceNodePortRange,omitempty" norman:"default=30000-32767"`
 	// Enabled/Disable PodSecurityPolicy
 	PodSecurityPolicy bool `yaml:"pod_security_policy" json:"podSecurityPolicy,omitempty"`
+	// Enable/Disable AlwaysPullImages admissions plugin
+	AlwaysPullImages bool `yaml:"always_pull_images,omitempty" json:"always_pull_images,omitempty"`
```
alena1108: I believe you can override the default list of admission controllers with the external arguments list. If you feel we have to include this plugin to the list of the default ones, we should just add it. And document on how to disable it.

client/management/v3/zz_generated_rancher_kubernetes_engine_config.go
```diff
@@ -32,6 +33,7 @@ type RancherKubernetesEngineConfig struct {
 	BastionHost         *BastionHost       `json:"bastionHost,omitempty" yaml:"bastionHost,omitempty"`
 	CloudProvider       *CloudProvider     `json:"cloudProvider,omitempty" yaml:"cloudProvider,omitempty"`
 	ClusterName         string             `json:"clusterName,omitempty" yaml:"clusterName,omitempty"`
+	DinD                bool               `json:"dind,omitempty" yaml:"dind,omitempty"`
```
galal-hussein: what is the reason behind adding this as an option in rkeconfig, what is the use case of using it without the cli?

apis/management.cattle.io/v3/rke_types.go
jgreat: In order to override the list you need to provide the whole list. This isn't easy to determine what's supposed to be in this list, nor convenient to set that option.  The AlwaysPullImages admission controller is recommended as a CIS Best practice, but I believe the downside is you'll end up not taking advantage of docker image cache on the host.

jgreat:
@alena1108 @galal-hussein please review https://github.com/rancher/types/pull/648 before this change. This change builds on top of changes explained there.

apis/management.cattle.io/v3/k8s_defaults.go
```diff
@@ -45,6 +44,7 @@ var (
 			},
 			Kubelet: map[string]string{
 				"tls-cipher-suites": "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305",
+				"cadvisor-port":     "0",
```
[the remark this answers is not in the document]
jgreat: I'm pretty sure metrics server is the replacement for cadvisor and not related.

apis/management.cattle.io/v3/rke_types.go
```diff
@@ -23,6 +23,8 @@ type RancherKubernetesEngineConfig struct {
 	Authorization AuthzConfig `yaml:"authorization" json:"authorization,omitempty"`
 	// Enable/disable strict docker version checking
 	IgnoreDockerVersion bool `yaml:"ignore_docker_version" json:"ignoreDockerVersion" norman:"default=true"`
+	// Is this a DinD install?
+	DinD bool `yaml:"dind,omitempty" json:"dind,omitempty"`
```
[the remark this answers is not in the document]
jgreat: This is really a part of #648 and related to https://github.com/rancher/rke/pull/1056

apis/management.cattle.io/v3/rke_types.go
alena1108: @jgreat the downside of making any parameter strongly typed is that:

1)  Exposing every k8s service param can be cumbersome as k8s has way too many
2)  It is not clear if this parameter would take precedence over custom admission controllers passed via external args.

I don't particularly mind for this particular parameter to be exposed as strongly typed though, but in the end it is your call.

apis/management.cattle.io/v3/k8s_defaults.go
alena1108: We did pull out the support; it is driven by https://github.com/rancher/types/blob/master/apis/management.cattle.io/v3/k8s_defaults.go#L18 for open source users, and SLA for the customers.

apis/management.cattle.io/v3/rke_types.go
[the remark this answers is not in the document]
alena1108: Based on the offline talk with @jgreat and @galal-hussein, we are going to support dind flag in rke internal cluster structure only, as it will be set using cli only.

apis/management.cattle.io/v3/k8s_defaults.go
jgreat: Removed 1.8 block

PR #652, opened 2018-12-17
Use slice instead of map for ingress paths

orangedeng:
Related Issue: https://github.com/rancher/rancher/issues/17065

### DIFF
--- a/apis/project.cattle.io/v3/schema/schema.go
+++ b/apis/project.cattle.io/v3/schema/schema.go
@@ -669,9 +669,6 @@ func addServiceOrDNSRecord(dns bool) types.SchemasInitFunc {
 
 func ingressTypes(schemas *types.Schemas) *types.Schemas {
 	return schemas.
-		AddMapperForType(&Version, v1beta1.HTTPIngressRuleValue{},
-			&m.SliceToMap{Field: "paths", Key: "path"},
-		).
 		AddMapperForType(&Version, v1beta1.HTTPIngressPath{},
 			&m.Embed{Field: "backend"},
 		).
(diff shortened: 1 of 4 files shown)


orangedeng:
@cjellick Logic updated, API won't break in the current version. We just sort the paths in reverse order

loganhz:
@gitlawr Please review it

gitlawr:
It looks good, but I wonder if we can put this mapper to `norman/types/mapper` as it looks generic.
cc @cjellick @ibuildthecloud

alena1108:
@cjellick there are no more breaking api changes, but still a decision to be made of where the sort method should go to - types or norman library

mapper/sort_map_slice.go
```diff
@@ -0,0 +1,55 @@
...
+	"sort"
+
+	"github.com/rancher/norman/types"
+	"github.com/rancher/norman/types/convert"
+)
+
+type SortMapSlice struct {
+	Field        string
+	SorterKey    string
+	ReverseOrder bool
```
cjellick: You only ever set reverOrder to true. Why over-complicate by adding this boolean?

loganhz:
@cjellick Do you mean we need to make a breaking API change? Change it from `map` to `array`?

orangedeng:
Updated.

cjellick:
@orangedeng The change looks fine, but it definitely breaks the UI. Open a UI issue and get a UI PR submitted and reviewed so that we can simultaneously merge the UI and the backend.
cc @loganhz

cjellick:
Waiting on UI to be done so I can coordinate merging, since this breaks the UI https://github.com/rancher/rancher/issues/17529

loganhz:
UI PR is ready for review and merge https://github.com/rancher/ui/pull/2585

cjellick:
LGTM

PR #661, opened 2018-12-21
Improving logging

aiwantaozi:
Problem:
current we don't have test func for logging target
current we not allow user to config send log to target in cluster
user want to exclude system log from cluster log

Solution:
add test func for logging target
allow user to config send log to target in cluster
add flag to include system log or not

Issue:
https://github.com/rancher/rancher/issues/15380
https://github.com/rancher/rancher/issues/15555
https://github.com/rancher/rancher/issues/13785


apis/management.cattle.io/v3/schema/schema.go
```diff
@@ -499,8 +499,44 @@ func logTypes(schema *types.Schemas) *types.Schemas {
 			m.DisplayName{}).
 		AddMapperForType(&Version, v3.ProjectLogging{},
 			m.DisplayName{}).
-		MustImport(&Version, v3.ClusterLogging{}).
-		MustImport(&Version, v3.ProjectLogging{})
+		MustImport(&Version, v3.ClusterTestInput{}).
+		MustImport(&Version, v3.ProjectTestInput{}).
+		MustImportAndCustomize(&Version, v3.ClusterLogging{}, func(schema *types.Schema) {
+			schema.ResourceActions = map[string]types.Action{
```
orangedeng: If we only support collection action, we should remove the resource action here.

apis/management.cattle.io/v3/schema/schema.go
```diff
@@ -499,8 +499,44 @@ func logTypes(schema *types.Schemas) *types.Schemas {
...
+				"test": {
+					Input: "clusterTestInput",
+				},
+				"dryRun": {
+					Input: "clusterTestInput",
+				},
+			}
+		}).
+		MustImportAndCustomize(&Version, v3.ProjectLogging{}, func(schema *types.Schema) {
+			schema.ResourceActions = map[string]types.Action{
```
orangedeng: If we only support collection action, we should remove the resource action here.

orangedeng:
LGTM

cjellick:
@aiwantaozi @orangedeng @loganhz this looks like changes that break API backwards compatibility to me. Is that an accurate assessment? If so, I want it to go through the design review process.

aiwantaozi:
Hi, this change include: 
1. change some fields from string to password
2. add two action test and dry-run
3. merge some fields to a loggingTarge but not change the API output
How do you think?

cjellick:
> merge some fields to a loggingTarge but not change the API output

does this mean that before the change, a resource in the API would look like:
```
{
  fieldA
  fieldB
}
```
and after the change, it looks like this:
```
{
   newField: {
     fieldA
     fieldB
   }
}
```
?

aiwantaozi:
@cjellick ```Merge some fields to a loggingTarge but not change the API output``` means we just change the go struct

Before:
```
type ClusterLoggingSpec struct {
	Elasticsearch
        Splunk
}
```

After:
```
type ClusterLoggingSpec struct {
	LoggingTargets
}

type LoggingTargets struct {
      Elasticsearch
      Splunk
      ...
}
```

cjellick:
OK. I get it now. :-)

LGTM. I'm doing a quick review of your rancher/rancher PR. It currently needs rebased again, but once you do that and I finish my review of that, I will merge this

aiwantaozi:
@cjellick we need to merge this system charts pr first
https://github.com/rancher/system-charts/pull/9

PR #664, opened 2019-01-02
Support local chart app upgrade & rollback

thxCode:
**Problem:**
- Cannot upgrade local chart app
- Cannot rollback local chart app

**Solution:**
- Add `Files` field to `AppUpgradeConfig`
- Add `Files` field to `AppRevisionStatus` to record local chart templates

**Issue:**
rancher/rancher#15711

### DIFF
--- a/apis/management.cattle.io/v3/catalog_types.go
+++ b/apis/management.cattle.io/v3/catalog_types.go
@@ -3,7 +3,6 @@ package v3
 import (
 	"github.com/rancher/norman/condition"
 	"github.com/rancher/norman/types"
-
 	"k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


gitlawr:
Why do we need the `ChartInfo` field. Given that it is only used for displaying info in cli, I think we can get that info from Chart.yml file in the `Files`

thxCode:
@gitlawr , I have removed `ChartInfo` field, using `App.Spec.ExternalID` `App.Status.AppRevisionID` -> `AppRevision` to get the info from `Chart.yaml`

gitlawr:
LGTM

cjellick:
LGTM

thxCode:
@cjellick , rebased

PR #665, opened 2019-01-02
Add Members field for RBAC for multiclusterapp and globalDNS

mrajashree:


apis/management.cattle.io/v3/multi_cluster_app.go
```diff
@@ -39,3 +40,8 @@ type Answer struct {
 	ClusterName string            `json:"clusterName,omitempty" norman:"type=reference[cluster]"`
 	Values      map[string]string `json:"values,omitempty" norman:"required"`
 }
+
+type Member struct {
+	Name       string `json:"name,omitempty" norman:"type=reference[user],required"`
+	AccessType string `json:"accessType,omitempty"`
```
alena1108: we have to make it enum

alena1108:
@mrajashree once we finalize the code path for multi cluster app rbac, we should implement similar thing for the global dns records.

apis/management.cattle.io/v3/multi_cluster_app.go
```diff
@@ -39,3 +40,8 @@ type Answer struct {
 	ClusterName string            `json:"clusterName,omitempty" norman:"type=reference[cluster]"`
 	Values      map[string]string `json:"values,omitempty" norman:"required"`
 }
+
+type Member struct {
+	UserName   string `json:"userName,omitempty" norman:"type=reference[user],required"`
```
alena1108: @cjellick suggested that if we are going to support groups, then we have to go with the approach similar to PRTB where we have userPrincipalID and groupPrincipalID

apis/management.cattle.io/v3/multi_cluster_app.go
```diff
@@ -39,3 +40,8 @@ type Answer struct {
...
+	AccessType string `json:"accessType,omitempty" norman:"type=enum,options=everything|readOnly|upgrade"`
```
alena1108: lets use `all` instead of `everything`

alena1108: And `update` instead of `upgrade` as it should cover update and rollback actions

alena1108:
LGTM

apis/management.cattle.io/v3/multi_cluster_app.go
```diff
@@ -39,3 +40,11 @@ type Answer struct {
 	ClusterName string            `json:"clusterName,omitempty" norman:"type=reference[cluster]"`
 	Values      map[string]string `json:"values,omitempty" norman:"required"`
 }
+
+type Member struct {
+	UserName           string `json:"userName,omitempty" norman:"type=reference[user]"`
```
alena1108: talked to @cjellick about the necessity to have both fields - *name and *principalName. Looks like we should be fine with just having *principalName

alena1108:
LGTM, will merge after the commits are squashed.

PR #666, opened 2019-01-02
upgrade rollback for multi cluster app 

kinarashah:
https://github.com/rancher/rancher/issues/16016#issuecomment-448095645


apis/management.cattle.io/v3/multi_cluster_app.go
```diff
@@ -39,3 +46,23 @@ type Answer struct {
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	TemplateVersionName string   `json:"templateVersionName,omitempty" norman:"type=reference[templateVersion],required"`
+	Answers             []Answer `json:"answers,omitempty"`
+}
+
+type MultiClusterAppUpgradeInput struct {
+	TemplateVersionName string   `json:"templateVersionName,omitempty"`
+	Answers             []Answer `json:"answers,omitempty"`
+	BatchSize           int      `json:"batchSize,omitempty"`
```
alena1108: we should define the upgrade strategy, similar to how k8s does it

apis/management.cattle.io/v3/multi_cluster_app.go
```diff
@@ -39,3 +46,23 @@ type Answer struct {
 	ClusterName string            `json:"clusterName,omitempty" norman:"type=reference[cluster]"`
 	Values      map[string]string `json:"values,omitempty" norman:"required"`
 }
+
+type MultiClusterAppRevision struct {
```
alena1108: lets make sure to sync with @mrajashree on the rbac. The same rbac rules should apply to the revision that apply to the app

apis/management.cattle.io/v3/multi_cluster_app.go
alena1108: Also we might want to set the strategy as the field on the MCA, instead of setting it as an action input. Similar to how k8s does. This way you don't have to pass it on upgrade and rollback actions

apis/management.cattle.io/v3/multi_cluster_app.go
```diff
@@ -39,3 +49,25 @@ type Answer struct {
...
 	Values      map[string]string `json:"values,omitempty" norman:"required"`
 }
+
+type UpgradeStrategy struct {
+	RollingUpdate *RollingUpdate `json:"rollingUpdate,omitempty"`
+}
+
+type RollingUpdate struct {
+	BatchSize int `json:"batchSize,omitempty"`
+	Timeout   int `json:"timeout,omitempty"`
```
alena1108: instead of a timeout, we need an interval

alena1108:
LGTM

PR #675, opened 2019-01-12
 Revamp cluster auth specific variables into struct

erikwilson:
Moves cluster spec variables needed for authentication into a new
LocalClusterAuthEndpoint structure, and defaults enabled to true.


apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -246,3 +243,9 @@ type RotateCertificateInput struct {
 	CACertificates bool     `json:"caCertificates,omitempty"`
 	Services       []string `json:"services,omitempty" norman:"type=enum,options=etcd|kubelet|kube-apiserver|kube-proxy|kube-scheduler|kube-controller-manager"`
 }
+
+type LocalClusterAuthEndpoint struct {
+	Enabled bool   `json:"enabled" norman:"default=true"`
```
ibuildthecloud: i really, really think this is supposed to be `*bool`

erikwilson: As per conversation, `default=true` has been removed so Enabled will default to false.

cjellick:
LGTM, but i dont want to merge until the rancher/rancher pr is ready

PR #685, opened 2019-01-24
Update catalog types for disk caching

erikwilson:
Add a new catalog condition for migrating catalog and templates to
remove files and questions, and new variables on template version for
easy retrieval from cache. Mark parameters that are injected with the
formatter as nocreate/noupdate. Adds comments for deprecated items.

For rancher/rancher/issues/14322


cjellick:
LGTM

PR #704, opened 2019-02-01
Set snapshot to false

moelsayed:
https://github.com/rancher/rancher/issues/17720

### DIFF
--- a/apis/management.cattle.io/v3/rke_types.go
+++ b/apis/management.cattle.io/v3/rke_types.go
@@ -197,13 +197,13 @@ type ETCDService struct {
 	Key string `yaml:"key" json:"key,omitempty"`
 	// External etcd prefix
 	Path string `yaml:"path" json:"path,omitempty"`
-	// Etcd Recurring snapshot Service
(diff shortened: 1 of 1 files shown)


apis/management.cattle.io/v3/rke_types.go
```diff
@@ -197,13 +197,13 @@ type ETCDService struct {
 	Key string `yaml:"key" json:"key,omitempty"`
 	// External etcd prefix
 	Path string `yaml:"path" json:"path,omitempty"`
-	// Etcd Recurring snapshot Service
-	Snapshot *bool `yaml:"snapshot" json:"snapshot,omitempty" norman:"default=true"`
+	// Etcd Recurring snapshot Service, used by rke only
+	Snapshot *bool `yaml:"snapshot" json:"snapshot,omitempty" norman:"default=false"`
```
alena1108: we have to take care of 2 things if we put this change in:

* make sure backup is set by default in the API store, if not passed via api/ui
* for legacy clusters, we have to make sure that the snapshot set to false won't reset the previous true value

alena1108:
LGTM

PR #707, opened 2019-02-05
GlobalDNS entry - Make FQDN non-updatable

prachidamle:
Fix for https://github.com/rancher/rancher/issues/17832

Editing FQDN of the DNS entry is not allowed, editing it does not get propagate to the external-dns provider. User can always add a new entry and delete the existing


prachidamle:
@alena1108 @kinarashah pls review

prachidamle:
Will be adding in another fix to handle updates to FQDN

PR #708, opened 2019-02-05
Update kontainer driver schemas

nathan-jenan-rancher:
This change adds a schema version field so that we can track what resource
version of a dynamic schema a schema was created from, to minimize the amount
of updates we need to do.

Issue:
https://github.com/rancher/rancher/issues/17712


erikwilson:
LGTM

cjellick:
LGTM

PR #715, opened 2019-02-12
Update negative field to positive field in cluster logging

aiwantaozi:
Problem:
excludeSystemComponent is a negative field

Solution:
change to includeSystemComponent and use true as default

Issue:
https://github.com/rancher/rancher/issues/15555


orangedeng:
LGTM

PR #717, opened 2019-02-12
Add TTL to GlobalDNS

prachidamle:
https://github.com/rancher/rancher/issues/17871


apis/management.cattle.io/v3/globaldns_types.go
```diff
@@ -19,6 +19,7 @@ type GlobalDNS struct {
 
 type GlobalDNSSpec struct {
 	FQDN                string   `json:"fqdn,omitempty" norman:"required"`
+	TTL                 string   `json:"ttl,omitempty"`
```
alena1108: we should probably change it to the actual type of the TTL, something like int. So UI/API can do the validation

prachidamle: changed.

alena1108:
LGTM

PR #718, opened 2019-02-12
Add GetUserByPrincipalID to user manager interface

mrajashree:


prachidamle:
Had following comments that dint show up here: We should expose the API with a better name than checkCache() and also the method should  get User either from cache and handle the case where it is not found in cache.

prachidamle:
LGTM

PR #722, opened 2019-02-14
add option inherit advance from alert group in the alert rule as the default value.

aiwantaozi:
Problem:
both alert group and alert rule advance option is confused, we need to
make then simple

Solution:
alert rule use alert group level advance option as default but user can
edit if they don't have to inherit from group

Issue:
https://github.com/rancher/rancher/issues/17538


aiwantaozi:
need to work with UI

orangedeng:
LGTM

apis/management.cattle.io/v3/alerting_types.go
```diff
@@ -195,6 +195,7 @@ type CommonGroupField struct {
 type CommonRuleField struct {
 	DisplayName string `json:"displayName,omitempty"`
 	Severity    string `json:"severity,omitempty" norman:"required,options=info|critical|warning,default=critical"`
+	IsInherited *bool  `json:"isInherited,omitempty" norman:"default=true"`
```
cjellick: should just be `inherited` not `isInherited`

loganhz:
Updated

cjellick:
lgtm

PR #729, opened 2019-02-20
change rke-tools for k8s 1.13 and 1.12

galal-hussein:


apis/management.cattle.io/v3/k8s_defaults.go
```diff
@@ -745,13 +745,13 @@ var (
 			IngressBackend:            m("k8s.gcr.io/defaultbackend:1.4"),
 			MetricsServer:             m("gcr.io/google_containers/metrics-server-amd64:v0.3.1"),
 		},
-		"v1.12.5-rancher1-1": {
+		"v1.12.5-rancher1-1-1": {
```
alena1108: add it as a new version

PR #730, opened 2019-02-20
Add k8s 1.12 and 1.13 to allow upgrade

galal-hussein:


apis/management.cattle.io/v3/k8s_defaults.go
```diff
@@ -819,6 +819,58 @@ var (
 			CoreDNS:                   m("coredns/coredns:1.2.6"),
 			CoreDNSAutoscaler:         m("gcr.io/google_containers/cluster-proportional-autoscaler-amd64:1.0.0"),
 		},
+		"v1.12.5-rancher1-1-1": {
```
alena1108: @galal-hussein can we put a comment that this is needed for 2.1, and why?

PR #731, opened 2019-02-21
GlobalDNSProvider root domain is now optional, also dont need GlobalDNSSystemImages

prachidamle:
1) The kubernetes-incubator/external-dns does not require rootdomain always set. 
Now it is also optional on Rancher side.

2) Since we are switching to use system-charts, we no longer need GlobalDNSSystemImages in types

Fixes:
https://github.com/rancher/rancher/issues/18322
https://github.com/rancher/rancher/issues/17900
https://github.com/rancher/rancher/issues/17559


alena1108:
RootDomain is also moved from the each provider to the spec level; need to coordinate the UI before merging

orangedeng:
LGTM

PR #734, opened 2019-02-24
Move RootDomain field from provider configs to generic spec

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/18346


alena1108:
LGTM

PR #745, opened 2019-02-28
Add field to save user's inherited roles from different targets

mrajashree:


prachidamle:
This is for  #18260

prachidamle:
LGTM

PR #758, opened 2019-03-06
Add multiarch support

loganhz:
This PR includes the following changes.

Update k8s 1.12.6 and 1.13.4 images to use docker manifest

Issue
rancher/rancher#16461


loganhz:
We need to run [this script](https://github.com/jianghang8421/publish-manifest/blob/master/publish-manifest-images.sh) to push the related manifest images

PR #763, opened 2019-03-12
Manual backup

moelsayed:
https://github.com/rancher/rancher/issues/18807

### DIFF
--- a/apis/management.cattle.io/v3/backup_types.go
+++ b/apis/management.cattle.io/v3/backup_types.go
@@ -14,7 +14,7 @@ const (
 )
 
 type BackupConfig struct {
-	// Enable Flag
+	// Enable or disable recurring backups in rancher
 	Enabled *bool `yaml:"enabled" json:"enabled,omitempty" norman:"default=true"`
 	// Backup interval in hours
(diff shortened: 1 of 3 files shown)


apis/management.cattle.io/v3/backup_types.go
```diff
@@ -51,6 +51,8 @@ type EtcdBackup struct {
 type EtcdBackupSpec struct {
 	// cluster ID
 	ClusterID string `json:"clusterId,omitempty" norman:"required,type=reference[cluster]"`
+	// manual backup flag
+	Manual bool `yaml:"manual" json:"manual,omitempty"`
```
alena1108: do we need this field on an api level? if it used internally, you might need to drop it in schema.

alena1108: Example of dropping a field: https://github.com/rancher/types/pull/665/files#diff-d040f71d732be0b270d23ab28a54e726R623

alena1108: Actually, even if not used, this field can be useful for the info purpose.

alena1108:
LGTM

PR #765, opened 2019-03-12
Add kube-api-auth image to k8s_defaults

erikwilson:
Needed for generated list of images for air-gap.

For rancher/rancher/issues/18755


alena1108:
LGTM

PR #775, opened 2019-03-28
Add HPA type to Rancher schema

orangedeng:
Related issue: https://github.com/rancher/rancher/issues/19084


mapper/merge_list_by_index.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	toObj, ok := data[m.To]
+	if !ok {
+		return
+	}
+	fromList := convert.ToMapSlice(fromObj)
+	toList := convert.ToMapSlice(toObj)
+	for i := 0; i < len(fromList) && i < len(toList); i++ {
+		fromItem := fromList[i]
+		toItem := toList[i]
```
gitlawr: If I understand correctly this mapper is for merging metric spec & status to the same entry. How do we guarantee that this `fromList[i]` and `toList[i]` is referencing the same metric?

apis/project.cattle.io/v3/schema/schema.go
```diff
@@ -1020,3 +1022,70 @@ func monitoringTypes(schemas *types.Schemas) *types.Schemas {
...
+			&m.AnnotationField{Field: "displayName"},
+			&m.DisplayName{},
+			&m.AnnotationField{Field: "description"},
+			&m.Embed{Field: "status"},
+			mapper.NewMergeListByIndexMapper("currentMetrics", "metrics", "type"),
+		).
+		AddMapperForType(&Version, autoscaling.MetricTarget{},
+			&mapper.IntOrStringParser{Field: "averageUtilization"},
+			&m.Enum{Field: "type", Options: []string{"Utilization", "Value", "AverageValue"}},
+			&m.Move{To: "utilization", From: "averageUtilization"},
```
gitlawr: Why do we want to trim the `average` in `averageUtilization`?

apis/project.cattle.io/v3/schema/schema.go
```diff
@@ -6,13 +6,14 @@ import (
...
-	"github.com/rancher/types/apis/project.cattle.io/v3"
+	v3 "github.com/rancher/types/apis/project.cattle.io/v3"
 	"github.com/rancher/types/factory"
 	"github.com/rancher/types/mapper"
 	"k8s.io/api/apps/v1beta2"
+	autoscaling "k8s.io/api/autoscaling/v2beta2"
 	batchv1 "k8s.io/api/batch/v1"
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
gitlawr: Unnecessary alias, and check the above one.

apis/project.cattle.io/v3/schema/schema.go
```diff
@@ -1020,3 +1022,70 @@ func monitoringTypes(schemas *types.Schemas) *types.Schemas {
...
+			mapper.CrossVersionObjectToWorkload{Field: "workloadId"},
+			&m.Required{Fields: []string{"workloadId", "maxReplicas"}},
+			&m.AnnotationField{Field: "displayName"},
+			&m.DisplayName{},
+			&m.AnnotationField{Field: "description"},
+			&m.Embed{Field: "status"},
+			mapper.NewMergeListByIndexMapper("currentMetrics", "metrics", "type"),
+		).
+		AddMapperForType(&Version, autoscaling.MetricTarget{},
+			&mapper.IntOrStringParser{Field: "averageUtilization"},
```
gitlawr: Why do we need the `IntOrStringParser` given that `averageUtilization` field is int

status/status.go
```diff
@@ -59,6 +59,8 @@ var transitioningMap = map[string]string{
 	"Updating":                    "updating",
 	"Waiting":                     "waiting",
 	"InitialRolesPopulated":       "activating",
+	"ScalingActive":               "pending",
+	"AbleToScale":                 "pending",
```
gitlawr: I don't find references for these statuses in this and the related Rancher PR, are they valid?

mapper/merge_list_by_index.go
orangedeng: It is from kubernetes code. https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/podautoscaler/horizontal.go#L240

apis/project.cattle.io/v3/schema/schema.go
orangedeng: Because we have options in type field and I want to keep the consistency
- type `Value` -> field `value`
- type `AverageValue` -> field `averageValue`
- type `Utilization` -> field `utilization`

status/status.go
orangedeng: There are generated.

orangedeng: It is valid. 
```
// ScalingActive indicates that the HPA controller is able to scale if necessary:
// it's correctly configured, can fetch the desired metrics, and isn't disabled.
ScalingActive HorizontalPodAutoscalerConditionType = "ScalingActive"

// AbleToScale indicates a lack of transient issues which prevent scaling from occurring,
// such as being in a backoff window, or being unable to access/update the target scale.
AbleToScale HorizontalPodAutoscalerConditionType = "AbleToScale"
```

orangedeng: This is for the UI display usage. If HPA rule is not ready, you can see the reason from UI.

orangedeng:
Updated

gitlawr:
LGTM

cjellick:
LGTM, but will there be a rancher/rancher that goes along with this (beyond just updating types)

orangedeng:
@cjellick Yes. We have rancher HPA PR too. https://github.com/rancher/rancher/pull/19246. Already linked to this PR in the beginning.

cjellick:
LGTM

PR #778, opened 2019-04-03
Expose additional GlobalDNSProvider params

prachidamle:
This is a PR to add additional parameters to configure external-dns provider
https://github.com/rancher/rancher/issues/19251

### DIFF
--- a/apis/management.cattle.io/v3/globaldns_types.go
+++ b/apis/management.cattle.io/v3/globaldns_types.go
@@ -52,13 +52,18 @@ type GlobalDNSProviderSpec struct {
 }
 
 type Route53ProviderConfig struct {
-	AccessKey string `json:"accessKey" norman:"notnullable,required,minLength=1"`
-	SecretKey string `json:"secretKey" norman:"notnullable,required,minLength=1,type=password"`
+	AccessKey       string `json:"accessKey" norman:"notnullable,required,minLength=1"`
+	SecretKey       string `json:"secretKey" norman:"notnullable,required,minLength=1,type=password"`
+	CredentialsPath string `json:"credentialsPath" norman:"default=/.aws"`
+	RoleArn         string `json:"roleArn,omitempty"`
+	Region          string `json:"region" norman:"default=us-east-1"`
+	ZoneType        string `json:"zoneType" norman:"default=public"`
 }
 
 type CloudflareProviderConfig struct {
-	APIKey   string `json:"apiKey" norman:"notnullable,required,minLength=1,type=password"`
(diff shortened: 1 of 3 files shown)


apis/management.cattle.io/v3/globaldns_types.go
```diff
@@ -52,13 +52,18 @@ type GlobalDNSProviderSpec struct {
 }
 
 type Route53ProviderConfig struct {
-	AccessKey string `json:"accessKey" norman:"notnullable,required,minLength=1"`
-	SecretKey string `json:"secretKey" norman:"notnullable,required,minLength=1,type=password"`
+	AccessKey       string `json:"accessKey" norman:"notnullable,required,minLength=1"`
+	SecretKey       string `json:"secretKey" norman:"notnullable,required,minLength=1,type=password"`
+	CredentialsPath string `json:"credentialsPath" norman:"default=/.aws"`
+	RoleArn         string `json:"roleArn"`
```
alena1108: shall we add omitempty to all the new values?

client/management/v3/zz_generated_cloudflare_provider_config.go
```diff
@@ -1,12 +1,14 @@
...
+	CloudflareProviderConfigFieldAPIKey       = "apiKey"
+	CloudflareProviderConfigFieldProxySetting = "proxySetting"
 )
 
 type CloudflareProviderConfig struct {
-	APIEmail string `json:"apiEmail,omitempty" yaml:"apiEmail,omitempty"`
-	APIKey   string `json:"apiKey,omitempty" yaml:"apiKey,omitempty"`
+	APIEmail     string `json:"apiEmail,omitempty" yaml:"apiEmail,omitempty"`
+	APIKey       string `json:"apiKey,omitempty" yaml:"apiKey,omitempty"`
+	ProxySetting string `json:"proxySetting,omitempty" yaml:"proxySetting,omitempty"`
```
alena1108: do we only want proxy parameter for cloudFlare? I recall seeing more params via answers, are any of them useful enough to be exposed?

prachidamle: No, rest of the parameters are not for configuring cloudflare functionality. They are mainly related to the external-dns pod deployment (use k8s rbac, ownerId to know which records are created by this pod, do a full-sync Vs only add new dns records etc.)

apis/management.cattle.io/v3/globaldns_types.go
prachidamle: I think we dont need omitempty for the ones which have a default specified. For the property "RoleArn" I will add omitempty.

alena1108:
LGTM

PR #789, opened 2019-04-09
 Add rancher min/max version to templateVersion

dramich:
Related: https://github.com/rancher/rancher/pull/19630


cjellick:
LGTM

PR #795, opened 2019-04-17
Add password tag to pwd/secrets fields for rke configs

mrajashree:
To address: https://github.com/rancher/rancher/issues/19653


alena1108:
LGTM

PR #798, opened 2019-04-20
Generate code for new init() method

daxmc99:
Wait to merge until https://github.com/rancher/norman/pull/277


cjellick:
code change for adding DynamicClient is good, but generated code will need updated based on changes to norman

cjellick:
LGTM

PR #801, opened 2019-04-23
Add service account fields to PRTB

orangedeng:
Add service account field to PRTB

Fields:
- ServiceAccountName
- ServiceAccountNamespace

Related issue: https://github.com/rancher/rancher/issues/19381

### DIFF
--- a/apis/management.cattle.io/v3/authz_types.go
+++ b/apis/management.cattle.io/v3/authz_types.go
@@ -123,6 +123,7 @@ type ProjectRoleTemplateBinding struct {
 	GroupPrincipalName string `json:"groupPrincipalName,omitempty" norman:"type=reference[principal]"`
 	ProjectName        string `json:"projectName,omitempty" norman:"required,type=reference[project]"`
 	RoleTemplateName   string `json:"roleTemplateName,omitempty" norman:"required,type=reference[roleTemplate]"`
+	ServiceAccount     string `json:"serviceAccount,omitempty" norman:"nocreate,noupdate"`
 }
 
 type ClusterRoleTemplateBinding struct {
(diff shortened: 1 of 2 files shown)


status/status.go
```diff
@@ -59,6 +59,8 @@ var transitioningMap = map[string]string{
 	"Updating":                    "updating",
 	"Waiting":                     "waiting",
 	"InitialRolesPopulated":       "activating",
+	"ServiceAccountCreated":       "creating",
+	"PrivilegesCreated":           "creating",
```
gitlawr: I think `Privilege` is not a proper word, we should change it to `Role` or `ServiceAccountRole`

orangedeng:
Updated

gitlawr:
LGTM

orangedeng:
Due to the offline discussion with @cjellick , I change to add service account fields into PRTB instead. No longer need to add field to app and app revision.
/cc @loganhz

apis/management.cattle.io/v3/authz_types.go
```diff
@@ -123,6 +123,9 @@ type ProjectRoleTemplateBinding struct {
 	GroupPrincipalName string `json:"groupPrincipalName,omitempty" norman:"type=reference[principal]"`
 	ProjectName        string `json:"projectName,omitempty" norman:"required,type=reference[project]"`
 	RoleTemplateName   string `json:"roleTemplateName,omitempty" norman:"required,type=reference[roleTemplate]"`
+
+	ServiceAccountName      string `json:"serviceAccountName,omitempty" norman:"nocreate,noupdate"`
```
cjellick: combine this and the ServiceAccountNamespace field into a single `:` separated field, so that it looks like `namespace:serviceAccountName`.

orangedeng:
Updated

alena1108:
LGTM

PR #807, opened 2019-04-26
Change CloudFlare proxySetting parameter to bool

prachidamle:
https://github.com/rancher/rancher/issues/19251


prachidamle:
@alena1108 review/merge pls.

apis/management.cattle.io/v3/globaldns_types.go
```diff
@@ -63,7 +63,7 @@ type Route53ProviderConfig struct {
 type CloudflareProviderConfig struct {
 	APIKey       string `json:"apiKey" norman:"notnullable,required,minLength=1,type=password"`
 	APIEmail     string `json:"apiEmail" norman:"notnullable,required,minLength=1"`
-	ProxySetting string `json:"proxySetting" norman:"default=true"`
+	ProxySetting bool   `json:"proxySetting" norman:"default=true"`
```
alena1108: default=true won't work for bool, as bool is false by default in go. If you want to make it true, it has to be *bool

PR #818, opened 2019-05-03
Add google oauth schema

mrajashree:


apis/management.cattle.io/v3/authn_types.go
```diff
@@ -162,6 +162,27 @@ type GithubConfigApplyInput struct {
 	Enabled      bool         `json:"enabled,omitempty"`
 }
 
+type GoogleOAuthConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	AuthConfig        `json:",inline" mapstructure:",squash"`
+
+	ClientCredentialsFile string `json:"clientCredentialsFile,omitempty" norman:"required"`
```
vincent99: Singular, not a file, should show up as `type: password`.  `googleKuberentesEngineConfig` just uses `credential`.

I don't remember exactly what the two different sets of credentials were for other than one was optional to make groups work, but these probably need better names.  Who's the client?  Where do I get `ClientCredentialsFile` and `ServiceAccountFile` from?

apis/management.cattle.io/v3/authn_types.go
```diff
@@ -162,6 +162,27 @@ type GithubConfigApplyInput struct {
...
+	ServiceAccountFile    string `json:"serviceAccountFile,omitempty"`
+	AdminEmail            string `json:"adminEmail,omitempty"`
```
vincent99: What admin?  Is this not already a part of one of the 2 credential JSON blobs above?

apis/management.cattle.io/v3/authn_types.go
```diff
@@ -162,6 +162,27 @@ type GithubConfigApplyInput struct {
...
+	Domain                string `json:"domain,omitempty" norman:"required"`
```
vincent99: `hostname`, as in `githubConfig`.

apis/management.cattle.io/v3/schema/schema.go
```diff
@@ -506,7 +506,25 @@ func authnTypes(schemas *types.Schemas) *types.Schemas {
...
 		MustImport(&Version, v3.SamlConfigTestInput{}).
-		MustImport(&Version, v3.SamlConfigTestOutput{})
+		MustImport(&Version, v3.SamlConfigTestOutput{}).
+		//GoogleOAuth Config
+		MustImportAndCustomize(&Version, v3.GoogleOAuthConfig{}, func(schema *types.Schema) {
+			schema.BaseType = "authConfig"
+			schema.ResourceActions = map[string]types.Action{
+				"disable": {},
+				"configureTest": {
+					Input:  "googleOAuthConfig",
```
vincent99: Lowercase `A` (`googleOauthConfig`) throughout.  At least in what's visible to the client, if go linting wants it capital in the Go types that's fine

apis/management.cattle.io/v3/authn_types.go
mrajashree: This is the admin of the hosted google domain. I've seen that it's required in making api calls when using service account (which is also created by this admin). And it's not a part of the above 2 json files

apis/management.cattle.io/v3/authn_types.go
mrajashree: Screen Shot 2019-05-28 at 12 33 14 PM
The client credentials file has the client ID and secret generated through the `Create Credentials` -> `OAuth Client ID` option above.
Once created, the download button on the right most side of a client id downloads a json file titled client_secret_<clientID>.json (That's why I named it ClientCredentialsFile)
So we can copy paste contents of this file/upload it directly as we do for saml providers’ cert/key/metadata. The golang-google-oauth library we use also accepts this json file when generating access token.
(Ref: https://developers.google.com/identity/protocols/OpenIDConnect)

Screen Shot 2019-05-28 at 1 14 41 PM
According to google docs this is how you can create a service account for domain delegation. And during creation it asks you to download a file containing a private key for the service account either in JSON or P12 format, and JSON is the recommended one. That's why I named this ServiceAccountFile
(Ref: https://developers.google.com/admin-sdk/directory/v1/guides/delegation)

Is this okay for the two files? or you think we should go with something else?

Yes I will change the type to password for both these, but I hadn't done this right now, since I'm using api ui to do all testing, and it becomes easier to input plaintext, so I was going to change it to password before it's merged

apis/management.cattle.io/v3/authn_types.go
mrajashree: This is the admin of the hosted google domain. I’ve seen that it’s required in making api calls when using service account (which is also created by this admin). And it’s not a part of the above 2 json files
Ref: https://developers.google.com/admin-sdk/directory/v1/guides/delegation

apis/management.cattle.io/v3/authn_types.go
vincent99: Yes;  there are no files in an API, so `File` implies that you're looking for a path to a file, which you are not.

I would call them `oauthCredential` and `serviceAccountCredential`, since they're both the same "credential" format.

apis/management.cattle.io/v3/authn_types.go
vincent99: Does the `client_email` in the credential work?  It doesn't really make sense to have to specify who the key is for again...

mrajashree: @vincent99 Even I thought it'd work but it didn't, I'll try another time to be sure, but even in the Go example they have given in this doc they use some other user email (admin of the domain) and not the client email.

apis/management.cattle.io/v3public/authn_types.go
```diff
@@ -37,11 +37,19 @@ type GithubProvider struct {
 	RedirectURL string `json:"redirectUrl"`
 }
 
-type GithubLogin struct {
+type CodeBasedLogin struct {
```
prachidamle: We are changing the Login Action Input, I guess this needs a UI change for Github provider? Do we have any integrations(automation test etc) other than UI that does github auth

prachidamle: Also I guess we need similar change for Azure AD provider?

mrajashree: Azure AD Login has it's own separate types schema https://github.com/rancher/types/blob/master/apis/management.cattle.io/v3public/authn_types.go#L69

Yeah maybe UI change will be required. I don't think we have any automation tests for Github. I think creating separate schema for google oauth login makes sense then like Azure AD's?

prachidamle: Yeah if our aim it to consolidate all "code" based logins, then we should also change AzureAD. But since it is already a v3public API, I guess its better to add new rather than affect existing auth providers

apis/management.cattle.io/v3/authn_types.go
```diff
@@ -162,6 +162,28 @@ type GithubConfigApplyInput struct {
...
+type GoogleOauthConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	AuthConfig        `json:",inline" mapstructure:",squash"`
+
+	OauthCredential          string `json:"oauthCredential,omitempty" norman:"required"`
+	ServiceAccountCredential string `json:"serviceAccountCredential,omitempty"`
+	AdminEmail               string `json:"adminEmail,omitempty"`
+	Hostname                 string `json:"hostname,omitempty" norman:"required"`
+	UserInfoEndpoint         string `json:"userInfoEndpoint" norman:"default=https://openidconnect.googleapis.com/v1/userinfo,required,notnullable"`
```
prachidamle: Do we need nocreate on this?

mrajashree: Discussed offline and decided it's not needed cc @alena1108 @prachidamle

apis/management.cattle.io/v3/authn_types.go
vincent99: I still don't know what this means... Presumably there can be more than one admin of the domain.  Does it have to be a particular one?  Any one?  The same as the one that's making the service account? 

What if they are no longer an admin or part of the domain tomorrow?

prachidamle:
LGTM

apis/management.cattle.io/v3/authn_types.go
mrajashree: The service account needs to impersonate a user that has access to the admin apis, hence the admin of the domain. So this is the admin user who is creating the service account.
```
Note: Only users with access to the Admin APIs can access the Admin SDK Directory API, therefore your service account needs to impersonate one of those users to access the Admin SDK Directory API. Additionally, the user must have logged in at least once and accepted the G Suite Terms of Service.
```

vincent99:
LGTM

PR #822, opened 2019-05-07
GlobalDNS FQDN should be of type hostname

prachidamle:
https://github.com/rancher/rancher/issues/19738

Now the GlobalDNS FQDN is of type "hostname" which is a series of dnsLabels, that can only have alphanumeric characters and hyphen(-), separated by dot(.)


leodotcloud:
LGTM

PR #828, opened 2019-05-08
GlobalDNS FQDN should be of type hostname

prachidamle:
Backporting to release/v2.2

https://github.com/rancher/rancher/issues/20088


alena1108:
LGTM

PR #834, opened 2019-05-23
Add fields to nodeDriverStatus

dramich:
Problem:
Nodedrivers don't hold info on what has been applied

Solution:
Add applied fields


cjellick:
lgtm

PR #835, opened 2019-05-23
make update false for immutable fields of persistentVolume

kinarashah:
https://github.com/rancher/rancher/issues/19237


alena1108:
LGTM

PR #836, opened 2019-05-23
fix openstack lb options from int to string

leodotcloud:
https://github.com/rancher/rancher/issues/14577


alena1108:
LGTM

rawmind0:
https://github.com/terraform-providers/terraform-provider-rancher2/pull/21

PR #831, opened 2019-05-15
add istio metrics graph types

fyery-chen:
`istiocluster` -- get istio metric information in cluster level
`istioproject` -- get istio metric information in project level

PR #839, opened 2019-05-28
add istio metrics graph types

orangedeng:
`istiocluster` -- get istio metric information in cluster level
`istioproject` -- get istio metric information in project level


loganhz:
https://github.com/rancher/types/pull/831#issuecomment-492565080

cjellick:
I believe we agreed that Istio would be just an app deployment, completely independent of Rancher. This tightly couples the app deployment to backend rancher code. It creates a dependency between Rancher and Istio that we had hoped to avoid.

We may need to revisit this design.

cjellick:
despite what i said above, this continues an existing paradigm, so i am willing to move forward with it. lgtm

PR #847, opened 2019-05-30
CIS changes

leodotcloud:


apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -30,6 +30,7 @@ const (
 	ClusterActionBackupEtcd            = "backupEtcd"
 	ClusterActionRestoreFromEtcdBackup = "restoreFromEtcdBackup"
 	ClusterActionRotateCertificates    = "rotateCertificates"
+	ClusterActionRunCISScan            = "runCisScan"
```
alena1108: @leodotcloud we have to change the action name

apis/management.cattle.io/v3/schema/schema.go
```diff
@@ -832,3 +834,7 @@ func clusterTemplateTypes(schemas *types.Schemas) *types.Schemas {
 		MustImport(&Version, v3.ClusterTemplate{}).
 		MustImport(&Version, v3.ClusterTemplateRevision{})
 }
+
+func clusterScanTypes(schemas *types.Schemas) *types.Schemas {
+	return schemas.MustImport(&Version, v3.ClusterScan{})
```
alena1108: we should disable create and update (PUT/POST) from the schema as it will be created internally only

alena1108:
LGTM

PR #855, opened 2019-06-10
Support set http_proxy config on Rancher notifier

aiwantaozi:
Problem:
Not support http_proxy setting now
Solution:
Add http_proxy configure for slack, pagerduty, wechat, webhook

Issue:
https://github.com/rancher/rancher/issues/17885


alena1108:
LGTM

PR #860, opened 2019-06-12
Update goimports

rmweir:
Problem:
goimports is outdated, causing builds to fail if machine that originally ran goimports is up to date. This is caused by the alias changes to goimports

Solution:
update goimports


ibuildthecloud:
👍lgtm

PR #868, opened 2019-06-14
Add selfHealing and nodeNotReadyTimeout to nodepool spec

daxmc99:
Enable self-healing by deleting nodes after they have been unreachable
for a configurable amount of time


cjellick:
lgtm

PR #872, opened 2019-06-17
ClusterTemplate use name mapper for displayName

prachidamle:
We should be mapping displayName -> name on the schema, similar to the Cluster object. UI does not expect the schema to expose displayName field.


alena1108:
LGTM

PR #880, opened 2019-06-21
Make service account creds required for google oauth

mrajashree:
In the absence of service account creds, non-admin users from google's
hosted domain cannot lookup groups, even their own groups. (Restriction
from google api). So even during login we can't find out what groups the 
user belongs to. Admins of hosted google domain do not have the same restriction.
So they can still search for and add groups that non-admin google users
belong to. This will cause lot of inconsistencies, since admins can add
a user's groups, but user still won't have access to rancher. So making
service account credentials a required field will avoid all this, and
searching users/groups will work equally for all google users.

Hostname should not be updatable once auth is setup, since the
principalIds assigned to users after they log into rancher contain
their google addresses which have the hostname.


alena1108:
LGTM

PR #883, opened 2019-06-25
Add taints support for node

orangedeng:
Add taints in rke node config
Add taints field for node pool
Add Windows prefer field in cluster because we don't have a specific field for windows preferred cluster.
So UI has to guess the cluster support windows node or not.

Related issues:
https://github.com/rancher/rancher/issues/20805
https://github.com/rancher/rancher/issues/13972


apis/management.cattle.io/v3/machine_types.go
```diff
@@ -193,8 +196,10 @@ type NodeSpec struct {
 	DisplayName              string            `json:"displayName"`
 	RequestedHostname        string            `json:"requestedHostname,omitempty" norman:"type=hostname,nullable,noupdate,required"`
 	InternalNodeSpec         v1.NodeSpec       `json:"internalNodeSpec"`
+	DesiredNodeTaints        []v1.Taint        `json:"desiredNodeTaints"`
 	DesiredNodeLabels        map[string]string `json:"desiredNodeLabels,omitempty"`
 	DesiredNodeAnnotations   map[string]string `json:"desiredNodeAnnotations,omitempty"`
+	UpdateTaintsFromAPI      *bool             `json:"updateTaintsFromAPI,omitempty" norman:"default=false"`
```
gitlawr: Can you describe how this field is used, and why it needs to be a pointer when it defaults to false

orangedeng: The default value is not good. Will be removed.
This field will be set to true by the store in Rancher. It means there are some changes from the api. 
And it will be set to false by the controller. It means there are some changes from the controller.
The linux taints controller won't update the `v3.Node` object if the `UpdateTaintsFromAPI` is not nil. It can avoid multiple changes in the same time and the second change overwrite the first change.

orangedeng:
Updated

gitlawr:
LGTM

apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -95,6 +95,7 @@ type ClusterSpecBase struct {
 	EnableNetworkPolicy                  *bool                          `json:"enableNetworkPolicy" norman:"default=false"`
 	EnableClusterAlerting                bool                           `json:"enableClusterAlerting" norman:"default=false"`
 	EnableClusterMonitoring              bool                           `json:"enableClusterMonitoring" norman:"default=false"`
+	WindowsPreferedCluster               bool                           `json:"windowsPreferedCluster" norman:"default=false,noupdate"`
```
alena1108: no need to set default=false; all booleans in go get defaulted to false w/o it

apis/management.cattle.io/v3/machine_types.go
```diff
@@ -42,10 +42,11 @@ type NodeTemplateCondition struct {
...
 type NodeTemplateSpec struct {
-	DisplayName         string `json:"displayName"`
-	Description         string `json:"description"`
-	Driver              string `json:"driver" norman:"nocreate,noupdate"`
-	CloudCredentialName string `json:"cloudCredentialName" norman:"type=reference[cloudCredential]"`
+	DisplayName         string     `json:"displayName"`
+	Description         string     `json:"description"`
+	Driver              string     `json:"driver" norman:"nocreate,noupdate"`
+	CloudCredentialName string     `json:"cloudCredentialName" norman:"type=reference[cloudCredential]"`
+	NodeTaints          []v1.Taint `json:"nodeTaints,omitempty"`
```
alena1108: is there a reason for adding taints, but not nodeLabels/nodeAnnotations, on the node pool? Usually they all come together at other places.

orangedeng: This is the requirement from @cjellick and @vincent99. They want to set node taints from node pool and node template. And set those taints for newly created node from node pool.

orangedeng: The node labels from the template are stored in `NodeTemplate.ObjectMeta.Labels` so the node label field don't need to defined in the `NodeTemplateSpec` struct.

apis/management.cattle.io/v3/cluster_types.go
orangedeng: Updated

orangedeng:
Rebased.

apis/management.cattle.io/v3/machine_types.go
alena1108: Yea, my question was more about "can we add similar support for labels/annotations on the pool". But if they get inherited from the object meta, and already working, perfect.

alena1108:
LGTM

PR #895, opened 2019-07-05
Bring back host field in HttpGet probe

gitlawr:
Address issues:
https://github.com/rancher/rancher/issues/20749
https://github.com/rancher/rancher/issues/19992

Problem:
On redeploy/UI update, httpGet.host in yaml manifest is removed.

Solution:
Bring it back in API scheme

The Host header in `HTTPHeaders` field can be a substitute for it, but we probably shouldn't break existing manifests using the `host` field.


orangedeng:
LGTM

cjellick:
This was intentionally and explicitly dropped by @ibuildthecloud in this pr https://github.com/rancher/types/pull/73. Unfortunately, not a good comment or linked issue.

We need to understand why he dropped it before we can add it back.

ibuildthecloud:
LGTM

PR #896, opened 2019-07-07
Template api changes

prachidamle:
Following changes are added:
- drop Enabled flag on ClusterTemplate. Enabled is present on ClusterTemplateRevision/
- create Cluster using ClusterTemplateRevision only, create cluster API does not require ClusterTemplateId


alena1108:
LGTM

PR #903, opened 2019-07-11
Update calico to 3.7.4 + add controllers image

superseb:
https://github.com/rancher/rancher/issues/21475


apis/management.cattle.io/v3/k8s_defaults.go
```diff
@@ -1253,11 +1253,12 @@ var (
 			KubeDNSAutoscaler:         m("gcr.io/google_containers/cluster-proportional-autoscaler:1.3.0"),
 			Flannel:                   m("quay.io/coreos/flannel:v0.11.0-rancher1"),
 			FlannelCNI:                m("rancher/flannel-cni:v0.3.0-rancher1"),
-			CalicoNode:                m("quay.io/calico/node:v3.7.3"),
-			CalicoCNI:                 m("quay.io/calico/cni:v3.7.3"),
+			CalicoNode:                m("quay.io/calico/node:v3.7.4"),
```
alena1108: we can't use change the existing version if it's been released already  - this version in particular will go to 2.2.5. As we are not planning to include calico fix to it due to qa time crunch, we'd need to introduce a new version of k8s to the list

PR #904, opened 2019-07-11
Optional timestamp on snapshots.

moelsayed:
https://github.com/rancher/rancher/issues/20819


apis/management.cattle.io/v3/backup_types.go
```diff
@@ -22,6 +22,8 @@ type BackupConfig struct {
 	Retention int `yaml:"retention" json:"retention,omitempty" norman:"default=6"`
 	// s3 target
 	S3BackupConfig *S3BackupConfig `yaml:",omitempty" json:"s3BackupConfig"`
+	// Timestamp is appended to snapshot names, defaults to false
+	Timestamp bool `yaml:"timestamp" json:"timestamp,omitempty"`
```
alena1108: We should make it default. Also we need a file name to be added to the backup so we can generate the name on the rancher side, and store it on etcd.

PR #907, opened 2019-07-16
Add ScalingLimited condition to status map

orangedeng:
The hpa condition ScalingLimited should be in reverseErrorMap.
It means the the hps object is currently not able to scale due to
the scaling speed is too high.

Related issue:
https://github.com/rancher/rancher/issues/20955


ibuildthecloud:
LGTM

PR #914, opened 2019-07-19
Add exist condition for expression alert

aiwantaozi:
Problem:
current we require the user to input a threshold number, could not support
the scenario any matched condition event happens

Solution:
add exist condition and remove the required in thresdhold

Issue:
https://github.com/rancher/rancher/issues/21302


gitlawr:
LGTM

luthermonson:
LGTM

cjellick:
lgtm

PR #915, opened 2019-07-22
Ingress name type to hostname

alena1108:
https://github.com/rancher/rancher/issues/20809


prachidamle:
how does this handle upgrade case for existing ingress resources? Since we are relaxing the type form "dnsLabel" to "hostName" - it should be ok I guess.

alena1108:
@prachidamle yes, upgrade should be fine as we are relaxing the rule.

prachidamle:
LGTM

PR #917, opened 2019-07-23
Add user types to process and etcd service

moelsayed:


superseb:
LGTM

apis/management.cattle.io/v3/rke_types.go
```diff
@@ -244,6 +244,11 @@ type ETCDService struct {
 	Key string `yaml:"key" json:"key,omitempty"`
 	// External etcd prefix
 	Path string `yaml:"path" json:"path,omitempty"`
+	// UID to run etcd container as
+	UID int `yaml:"uid" json:"uid,omitempty"`
```
alena1108: can you provide an example of gid/uid

alena1108: We'll need to mention it in the steps for qa as well. Also can you describe why we can't use a string type here

moelsayed: On a typical ubuntu server:
```
$ id ubuntu 
uid=1000(ubuntu) gid=1000(ubuntu)
```
However, for this particular use case, the user will need to create a specific user for etcd on the node before provisioning:
```
# useradd etcd 
# id etcd
uid=1001(etcd) gid=1001(etcd) groups=1001(etcd)
```

The string username/group are just user friendly representations of the actual integer uid/gid used by the system, the mapping is done in the /etcd/passed file. If we use  username/group in this case, docker will fail to find them in the etcd docker container /etcd/passed file, and will fail to run the container. However, if we use an integer value, docker will be able use it and run the container even if that uid/gid doesn't map to an actual user inside the container. That's why we can only use the integer values here.

alena1108:
LGTM

PR #920, opened 2019-07-25
Update defaultRecipient of slackConfig to not required

guangbochen:
Update DefaultRecipient of SlackConfig to not required

**Related Issue:**
https://github.com/rancher/rancher/issues/20067

### DIFF
--- a/apis/management.cattle.io/v3/alerting_types.go
+++ b/apis/management.cattle.io/v3/alerting_types.go
@@ -289,7 +289,7 @@ type SMTPConfig struct {
 }
 
 type SlackConfig struct {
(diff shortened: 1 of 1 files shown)


aiwantaozi:
The slack webhook could be used for all channel, which means it doesn't matter whether you configure channel A or B, the webhook URL is the same, so when you configure channel A in slack, you can also configure B in Rancher, we will send alert to B, so the default recipient is required like others notifiers. It's a UI/UX problem. If we really need to remove this, we need more test, alertmanager may need this field to start, if alertmanager need it, and we don't let the user input it, we may need to check slack API allow us to get it.

loganhz:
LGTM

PR #923, opened 2019-07-27
Fix marshal v1.Taint to yaml panic

orangedeng:
**Problem:**
Marshal v1.Taint with timeAdded nil value to yaml panic.

**Solution:**
Use our own taint struct with omitempty flag of timeAdded field.


loganhz:
LGTM

PR #929, opened 2019-08-01
Remove scaling limited from error map

orangedeng:
**Problem:**
The `ScalingLimited` will set to true when the target workload's
cpu useage is zero.
In this case, we should not set state to error.

**Solution:**
Remove `ScalingLimited` from reverseErrorMap.

Related issue: 
https://github.com/rancher/rancher/issues/20955


loganhz:
LGTM

luthermonson:
LGTM

cjellick:
lgtm

PR #930, opened 2019-08-02
Add storage client

rmweir:
Problem:
Cannot interact with k8s storage API without using k8s api directly.

Solution:
Add generate client for storage api.

Related issue:
https://github.com/rancher/rancher/issues/20315


dramich:
lgtm

PR #936, opened 2019-08-07
Use norman struct tags for limits of nodepool timeout

daxmc99:
Extreme values could overflow nodepool timeout int64 value. Now, it will
return an error for values > maxInt64 or less than 0.

Addresses: https://github.com/rancher/rancher/issues/21431


rmweir:
lgtm

dramich:
lgtm

PR #937, opened 2019-08-09
Modify norman struct tag to be 1 year in seconds

daxmc99:
Javascript only supports a max int of 2^53 so very large values
can cause errors.

Instead, use the max value of a year in seconds for node timeout limit

Addresses: https://github.com/rancher/rancher/issues/21431


luthermonson:
LGTM

PR #942, opened 2019-08-16
Add version to test rke-tools v0.16.0-1 upgrade

superseb:


deniseschannon:
lgtm

deniseschannon:
This k8s version is noted as just for testing for the rke-tools update, but will need to be updated before we release.

PR #943, opened 2019-08-16
Add version to test rke-tools v0.1.41

superseb:


deniseschannon:
lgtm

These k8s version are just for testing for the rke-tools update, but will need to be updated before we release the next 2.2

PR #946, opened 2019-08-16
Add PutClusterScoped to NewLifecycleAdapter and remove from controllers

daxmc99:


rmweir:
I think updating/trashing norman and running go generate should be in separate commits

rmweir:
lgtm

PR #958, opened 2019-08-22
Add SendResolved field to Notifier

gitlawr:
issue: https://github.com/rancher/rancher/issues/17962

`send_resolved` is a common configuration for all kinds of notifier type.
https://prometheus.io/docs/alerting/configuration

It is set to false before.


aiwantaozi:
LGTM

PR #961, opened 2019-08-23
Add flag to disable snapshot timestamps

moelsayed:
https://github.com/rancher/rancher/issues/22019


leodotcloud:
@moelsayed Instead of disabling the timestamps, why don't you change the format? 
`time.Now().UnixNano())` or the equivalent?

alena1108:
LGTM

leodotcloud:
Summary of the offline discussion:
- We would like to use a standard format for the timestamps, hence we will stick with the current one (`time.RFC3339`)
- `time.Now().UnixNano())` provides a good alternative, but it's not human readable.
- Introducing the safe flag strikes a balance and also not break existing installs

leodotcloud:
LGTM

PR #972, opened 2019-09-05
Docker secret auth

prachidamle:
- Mapper should be invoked for v1.Secret type so that the "auth" field gets generated for dockerCedential
- Support "email" field in DockerCredential

### DIFF
--- a/apis/project.cattle.io/v3/schema/registry_credential.go
+++ b/apis/project.cattle.io/v3/schema/registry_credential.go
@@ -2,6 +2,7 @@ package schema
 
 import (
 	"encoding/base64"
+	"fmt"
 
 	"github.com/rancher/norman/types"
 	"github.com/rancher/norman/types/convert"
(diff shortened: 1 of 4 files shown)


apis/project.cattle.io/v3/schema/registry_credential.go
```diff
@@ -18,17 +19,33 @@ func (e RegistryCredentialMapper) ToInternal(data map[string]interface{}) error
...
+func addAuthInfo(data map[string]interface{}) error {
+
+	registryMap := convert.ToMapInterface(data["registries"])
+	for _, regCred := range registryMap {
+		regCredMap := convert.ToMapInterface(regCred)
+
+		username := convert.ToString(regCredMap["username"])
+		password := convert.ToString(regCredMap["password"])
+
+		if username == "" || password == "" {
```
alena1108: Nit: we can check the username right after line 37, and return if empty. this way we save on the conversion call for the password

apis/project.cattle.io/v3/schema/registry_credential.go
```diff
@@ -18,17 +19,35 @@ func (e RegistryCredentialMapper) ToInternal(data map[string]interface{}) error
 		return nil
 	}
 
-	auth := convert.ToString(data["auth"])
-	username := convert.ToString(data["username"])
-	password := convert.ToString(data["password"])
-
-	if auth == "" && username != "" && password != "" {
-		data["auth"] = base64.StdEncoding.EncodeToString([]byte(username + ":" + password))
+	if data["kind"] != "dockerCredential" {
```
kinarashah: can this mapper be added only on dockerCredential in schema?

prachidamle: It was not getting invoked for v3.DockerCredential

PR #973, opened 2019-09-06
add displayName to cloud credential 

kinarashah:


mapper/credentialmapper.go
```diff
@@ -7,17 +7,39 @@ import (
...
+		id := convert.ToString(values.GetValueN(data, "id"))
+		if id != "" {
+			data["name"] = id
+		}
+	}
 	delete(data, "data")
 }
 
 func (s CredentialMapper) ToInternal(data map[string]interface{}) error {
+	logrus.Infof("toInternal %v", data)
```
leodotcloud: Did you want to make it `Debugf` or remove it?

kinarashah: added it for my own logs, removed it

apis/management.cattle.io/v3/machine_types.go
```diff
@@ -319,5 +319,6 @@ type CloudCredential struct {
 }
 
 type CloudCredentialSpec struct {
+	DisplayName string `json:"displayName"`
```
prachidamle: Should this have norman:required?

kinarashah: @prachidamle we didn't have validation on this before, so I am not sure if adding it now would break anything for people who automate things and don't add anything. what do you think?

prachidamle: oh, if name is not required to create credential, we dont need displayName as required.

mapper/credentialmapper.go
```diff
@@ -9,15 +9,35 @@ import (
 	"github.com/rancher/norman/types/values"
 )
 
+const credAnn = "io.cattle.cloud.cred_name"
```
prachidamle: Do we have any format for adding annotations? I see some annotations with "field.cattle.io/" prefix - should we be following that?

kinarashah: I see that using `field.cattle.io` in annotations has its own mapper, mapper/annotation_field.go and I don't think that logic applies here .. I named it similar to `io.cattle.kontainer_driver.name` (which is label but used for internal purposes)

PR #976, opened 2019-09-13
set name annotation if name is empty

kinarashah:
Issue: setting name doesn't work, because annotation_field mapper overrides the name with field.cattle.io/name value

https://github.com/rancher/rancher/issues/22832


prachidamle:
LGTM

PR #983, opened 2019-09-23
Change the json-tag of WindowsPodInfraContainer in RKESystemImages 

thxCode:
**Problem:**
On air-gapped, the private registry could not apply to the pause image
of Windows kubelet. It seems the decoding of [mapstructure](github.com/mitchellh/mapstructure) doesn't
respect the json tag of the struct definition:
https://github.com/rancher/rancher/blob/0ba3fd34cc5bbbfda74449daa3b864f436a9f0ec/pkg/controllers/management/clusterprovisioner/provisioner.go#L751-L764

**Solution:**
Correct the json-tag

**Issue:**
https://github.com/rancher/rancher/issues/23000

### DIFF
--- a/apis/management.cattle.io/v3/rke_types.go
+++ b/apis/management.cattle.io/v3/rke_types.go
@@ -147,7 +147,7 @@ type RKESystemImages struct {
 	// Metrics Server image
 	MetricsServer string `yaml:"metrics_server" json:"metricsServer,omitempty"`
 	// Pod infra container image for Windows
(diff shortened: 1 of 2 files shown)


loganhz:
lgtm

thxCode:
After this merged, we need to update the vendors of the following repos:
- `rancher/kontainer-driver-metadata`, update `rancher/types` vendor, generate new data
- `rancher/rke`, update both `rancher/types` and `rancher/kontainer-driver-metadata`
- `rancher/rancher`, update `rancher/types`, `rancher/kontainer-driver-metadata` and `rancher/rke`

PR #987, opened 2019-09-25
RFC/PREP: Allow a load-balancer address for k8s API

rowanj:
Extends the RancherKubernetesEngineConfig (cluster.yml) type to support the configuration of an API load-balancer.

This is a pre-requisite for an upcoming PR to extend RKE to support such a configuration.

to-do:

- [ ] update and document design in accordance w/ feedback from #1348

--

This optional string is to replace the hostname used by other nodes and
configurations when attempting contacting the k8s API service.

It is expected to be the hostname of a layer-4 (TCP) load-balancer which
will be externally configured by the organisation to direct traffic to
an appropriate selection of the cluster's ControlPlane nodes on the same
port as it arrived; i.e. TCP/6443 for kube-apiserver.

Any TLS configuration and healthchecks on the load-balancer host are
also expected to be managed by the organisation.

### DIFF
--- a/apis/management.cattle.io/v3/rke_types.go
+++ b/apis/management.cattle.io/v3/rke_types.go
@@ -47,6 +47,8 @@ type RancherKubernetesEngineConfig struct {
 	AddonJobTimeout int `yaml:"addon_job_timeout" json:"addonJobTimeout,omitempty" norman:"default=30"`
 	// Bastion/Jump Host configuration
 	BastionHost BastionHost `yaml:"bastion_host" json:"bastionHost,omitempty"`
+	// Load Balancer address (externally-managed) through which nodes are to contact the control plane API port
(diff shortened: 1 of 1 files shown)


superseb:
The design for a solution for this has been posted at https://github.com/rancher/rke/pull/1348#issuecomment-527900505 and will be put in a final design soon that can be worked on. Is this taken into account or is this an implementation that you thought of? And is that design described somewhere?

rowanj:
Need to regenerate the built code but the build seems to be failing here due to our intercepting proxy; will continue from home.

rowanj:
@superseb I didn't see that PR in my search, this is based on our current need

rowanj:
I like most of the suggested design there, I'll adapt and extend and update this PR.

superseb:
There is now accepted design yet, I just opened https://github.com/rancher/rke/issues/1682, will have it reviewed and we can work from there.

PR #988, opened 2019-09-26
Update README.md

daxmc99:
For go1.13 and beyond we next to explicitly disable go modules since it will be enabled for all go commands


dramich:
lgtm

PR #1001, opened 2019-10-10
2.2 version changes

mrajashree:


alena1108:
We should make 2 commits - base copy and changes. This way it's easier to review

prachidamle:
LGTM
@mrajashree we also need a corresponding:
1) rke PR with types vendored in
2) rancher PR with types and rke vendored in

alena1108:
LGTM

PR #1007, opened 2019-10-15
[forwardport] Respect natively set host port

alena1108:
https://github.com/rancher/rancher/issues/23375

### DIFF
--- a/mapper/container_ports.go
+++ b/mapper/container_ports.go
@@ -6,6 +6,7 @@ import (
 	"github.com/rancher/norman/types"
 	"github.com/rancher/norman/types/convert"
 	"github.com/rancher/norman/types/mapper"
+	"github.com/rancher/norman/types/values"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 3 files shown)


mapper/container_ports.go
```diff
@@ -20,26 +21,65 @@ func (n ContainerPorts) FromInternal(data map[string]interface{}) {
...
+			portMap := map[string]interface{}{}
 			for _, port := range portsSlice {
 				asMap, err := convert.EncodeToMap(port)
 				if err != nil {
 					logrus.Warnf("Failed to convert container port to map %v", err)
 					continue
 				}
 				asMap["type"] = "/v3/project/schemas/containerPort"
-				containerPorts = append(containerPorts, asMap)
+				name, _ := values.GetValue(asMap, "name")
```
prachidamle: Will there always be a "name" for the port spec?

mapper/container_ports.go
```diff
@@ -20,26 +21,65 @@ func (n ContainerPorts) FromInternal(data map[string]interface{}) {
...
 
-	for i := 0; i < len(ports) && i < len(containers); i++ {
+	// process fields defined via annotations
+	for i := 0; i < len(annotationPorts) && i < len(containers); i++ {
 		container := convert.ToMapInterface(containers[i])
 		if container != nil {
-			portsSlice := convert.ToInterfaceSlice(ports[i])
-			var containerPorts []interface{}
+			portsSlice := convert.ToInterfaceSlice(annotationPorts[i])
+			portMap := map[string]interface{}{}
```
prachidamle: Is it always that the "containers[i]" will be associated with annotationPorts[i] ?

mapper/container_ports.go
alena1108: No, it can be empty for the case when it's just one port. For the case where there is more than one port, the name will always be required.

mapper/container_ports.go
alena1108: yes

prachidamle:
LGTM

PR #1012, opened 2019-10-25
New action on cluster saveAsTemplate

deniseschannon:
Note: Istio gateway is already in `release/v2.3`, which is why it isn't added again.


deniseschannon:
https://github.com/rancher/rancher/issues/22992

PR #1013, opened 2019-10-25
Add valid kubelet serving certificate

superseb:
```
services:
  kubelet:
    generate_serving_certificate: true
```


alena1108:
LGTM

PR #1023, opened 2019-11-05
Respect natively set host port

alena1108:
https://github.com/rancher/rancher/issues/23375


alena1108:
Corresponding forward port request is LGTMed https://github.com/rancher/types/pull/1007

PR #1024, opened 2019-11-06
Accept extra env vars, volume and volumeMounts for ingress addon

mrajashree:
Accept extra env vars, volume and volumeMounts for ingress addon

K8s supports different types of volumes, and the structs to define these have varying fields. So to accept volumes/volumeMounts from users, these fields refer the native k8s types.
These fields will then be parsed into the ingress template through rke.
The fields are named as per this comment to match the ones in nginx helm chart https://github.com/rancher/rke/issues/1104#issuecomment-547731864

https://github.com/rancher/rke/issues/1104


apis/management.cattle.io/v3/rke_types.go
```diff
@@ -400,6 +400,12 @@ type IngressConfig struct {
 	ExtraArgs map[string]string `yaml:"extra_args" json:"extraArgs,omitempty"`
 	// DNS Policy
 	DNSPolicy string `yaml:"dns_policy" json:"dnsPolicy,omitempty"`
+	// Extra Env vars
+	ExtraEnvs []v1.EnvVar `yaml:"extra_envs" json:"extraEnvs,omitempty" norman:"type=array[json]"`
+	// Extra volumes
+	ExtraVolumes []v1.Volume `yaml:"extra_volumes" json:"extraVolumes,omitempty" norman:"type=array[json]"`
```
alena1108: @mrajashree can you explore if it's possible to expose the field in our api as a concrete type []v1.Volume vs array[json]

mrajashree: As per offline discussion, sticking to array[json] so that the "Edit as YAML" does not add unset fields with default nil values.
But we'll use embedded types so that when we import them in schema.go we get a v3/schema type for each, so users can refer that for input.
And there is custom api level schema validation check on rancher side for these fields

PR #1025, opened 2019-11-06
Use default mapper for hostport conversion

alena1108:
To address https://github.com/rancher/rancher/issues/23375#issuecomment-550438582

* For hostPort port type, the sourcePort->hostPort conversion will be done by default mapper (as a reverse for https://github.com/rancher/types/blob/master/apis/project.cattle.io/v3/schema/schema.go#L517)
* For any other port, sourcePort will be dropped https://github.com/rancher/types/compare/release/v2.3...alena1108:nov6?expand=1#diff-8d858d11bc93015c62cc81c1f203550bR103 so the conversion doesn't happen

### DIFF
--- a/mapper/container_ports.go
+++ b/mapper/container_ports.go
@@ -98,13 +98,11 @@ func (n ContainerPorts) ToInternal(data map[string]interface{}) error {
 					logrus.Warnf("Failed to encode port: %v", err)
 					return obj
 				}
-				if strings.EqualFold(convert.ToString(mapped["kind"]), "HostPort") {
-					if _, ok := mapped["sourcePort"]; ok {
-						mapped["hostPort"] = mapped["sourcePort"]
-					}
+				if !strings.EqualFold(convert.ToString(mapped["kind"]), "HostPort") {
+					// delete the source port so it doesn't get converted to the host port by default mapper
+					delete(mapped, "sourcePort")
 				}
-				// delete the source port so it doesn't get converted to the host port by default mapper
-				delete(mapped, "sourcePort")
+
 			}
 			ports = append(ports, l)
 		}


mapper/container_ports.go
```diff
@@ -98,13 +98,11 @@ func (n ContainerPorts) ToInternal(data map[string]interface{}) error {
 					logrus.Warnf("Failed to encode port: %v", err)
 					return obj
 				}
-				if strings.EqualFold(convert.ToString(mapped["kind"]), "HostPort") {
-					if _, ok := mapped["sourcePort"]; ok {
-						mapped["hostPort"] = mapped["sourcePort"]
-					}
+				if !strings.EqualFold(convert.ToString(mapped["kind"]), "HostPort") {
+					// delete the source port so it doesn't get converted to the host port by default mapper
+					delete(mapped, "sourcePort")
```
prachidamle: Is there a possibility of losing the information for nodePort/LB since we will not be saving the port info to the etcd? In case there is a delay in the controllers launching  the corresponding nodeport/LB services?

mapper/container_ports.go
```diff
@@ -98,13 +98,11 @@ func (n ContainerPorts) ToInternal(data map[string]interface{}) error {
...
 				}
-				// delete the source port so it doesn't get converted to the host port by default mapper
-				delete(mapped, "sourcePort")
+
```
prachidamle: should we also have an integration test to check this usecase of non-native hostport is preserved?

alena1108: good idea @prachidamle

mapper/container_ports.go
alena1108: no @prachidamle. It gets preserved in "port" annotation.

mapper/container_ports.go
alena1108: @prachidamle added the test https://github.com/rancher/rancher/pull/23974

PR #1036, opened 2019-11-19
Make updateStrategy and replicas configurable for addons

mrajashree:
https://github.com/rancher/rancher/issues/23907

This commit makes update strategy configurable for all 4 system addons: ingress, networking, monitoring and dns
The ingress and networking addons are launched as daemonsets. If user doesn't pass any value for their update strategy, kubernetes sets default update strategy to rollingUpdate with maxUnavailable=1
The dns and monitoring addons are launched as deployments. If user doesn't pass any value for their update strategy, kubernetes sets default update strategy to rollingUpdate with maxUnavailable=25% and maxSurge=25%

The replicas field for monitoring deployment is made user configurable and will default to 2.
The dns addons use cluster-proportional-autoscaler which determine their replicas. So for dns, the fields needed for cluster-proportional-autoscaler are made configurable.


mrajashree:
@superseb I don't think this will affect existing clusters, since it needs rke&k-d-m template changes to pass in these values

apis/management.cattle.io/v3/rke_types.go
```diff
@@ -804,6 +809,10 @@ type MonitoringConfig struct {
 	Options map[string]string `yaml:"options" json:"options,omitempty"`
 	// NodeSelector key pair
 	NodeSelector map[string]string `yaml:"node_selector" json:"nodeSelector,omitempty"`
+	// Update strategy
+	UpdateStrategy *appsv1.DeploymentStrategy `yaml:"update_strategy" json:"updateStrategy,omitempty"`
+	// Number of monitoring addon pods
+	Replicas *int32 `yaml:"replicas" json:"replicas,omitempty" norman:"default=2"`
```
superseb: This is a change from the current config, why was configured with 2 replicas?

mrajashree: Having more than one replica will be good for zero-downtime use case when the metrics-server deployment undergoes rolling update.

PR #1043, opened 2019-12-02
update notifier fields type

aiwantaozi:
Problem: Notifier email configure return plaintext password
Solution: update type to password
Issue: https://github.com/rancher/rancher/issues/24335


loganhz:
LGTM

alena1108:
@aiwantaozi we have to make sure to:

1) add a test
2) validate that the subsequent update use case will work, from both the test + API. There were issues in the past when the password was reset on the crd when the UI was setting an empty value. Steps to take:
* Setup notifier with the password. Make sure the password is set on crd, and not shown in api
* Edit notifier. Change something non-password related, hit save. Make sure that the password in crd remains the same

loganhz:
Close in favor of https://github.com/rancher/types/pull/1051/files

PR #1051, opened 2019-12-09
Update notifier fields type

XianglongLuo:
Problem: Notifier email configure return plaintext password
Solution: update type to password
Issue: https://github.com/rancher/rancher/issues/24335
Related PR: https://github.com/rancher/rancher/pull/24477


loganhz:
Hi @alena1108 

```
we have to make sure to:

1) add a test
2) validate that the subsequent update use case will work, from both the test + API. There were issues in the past when the password was reset on the crd when the UI was setting an empty value. Steps to take:
* Setup notifier with the password. Make sure the password is set on crd, and not shown in api
* Edit notifier. Change something non-password related, hit save. Make sure that the password in crd remains the same
```

We have added the integration test for it in https://github.com/rancher/rancher/pull/24477

PR #1054, opened 2019-12-12
CIS fixes

leodotcloud:
- Change skip format
- Introduce override Benchmark version option
- Add more conditions


apis/management.cattle.io/v3/cluster_scan_types.go
```diff
@@ -24,7 +27,9 @@ const (
 
 type CisScanConfig struct {
 	// IDs of the checks that need to be skipped in the final report
-	Skip []string `json:"skip"`
+	Skip string `json:"skip"`
```
alena1108: omitempty flag?

apis/management.cattle.io/v3/cluster_scan_types.go
```diff
@@ -24,7 +27,9 @@ const (
...
+	// Override the CIS benchmark version to use for the scan (instead of latest)
+	OverrideBenchmarkVersion string `json:overrideBenchmarkVersion`
```
alena1108: same here

apis/management.cattle.io/v3/cluster_scan_types.go
leodotcloud: 👍

PR #1057, opened 2019-12-14
[forwardport] add AgentImageOverride

luthermonson:
https://github.com/rancher/rancher/issues/21386


dramich:
LGTM

cjellick:
@kinarashah @mrajashree @dramich As a heads up: I've asked @luthermonson to move AgentImageOverride off of clusterStatus and onto clusterSpec. It is a field that a user should set and thus should not be on the status object.

PR #1065, opened 2019-12-27
Use correct defaultbackend image, add image checks

superseb:
https://github.com/rancher/rancher/issues/24774


image/mirror.go
```diff
@@ -1,20 +1,23 @@
...
+	"fmt"
+	"strings"
+)
 
 var Mirrors = map[string]string{}
 
 func Mirror(image string) string {
 	orig := image
-	if strings.HasPrefix(image, "weaveworks") {
+	if strings.HasPrefix(image, "weaveworks") || strings.HasPrefix(image, "registry") || strings.HasPrefix(image, "busybox") {
```
prachidamle: Is there a ticket for this fix? Want to understand the usecase for returning image name as-is in these additional cases

image/mirror.go
```diff
@@ -25,6 +28,14 @@ func Mirror(image string) string {
...
 	image = strings.Replace(image, "minio/", "rancher/minio-", 1)
 
+	if !strings.HasPrefix(image, "rancher/") {
+		panic("image not prefixed with rancher/")
+	}
+
+	imageNameTag := strings.Split(image, ":")
+	if strings.HasSuffix(imageNameTag[0], "-") {
+		panic(fmt.Sprintf("Image [%s] has trailing '-', probably an error in image subsitution", image))
+	}
```
prachidamle: I dont think we should be adding checks here and panic on failure - There is another API for kontainer-driver-metadata @kinarashah has added that calls this - that API handler needs to be able to handle the error and add suitable response.

image/mirror.go
superseb: We have an exclude in the code for `weaveworks` but not for the others that we don't do any string manipulation, this change makes it so people reading the code understand what we don't manipulate.

image/mirror.go
superseb: This makes sure that the resulting images are valid, there is no case to proceed if the resulting images are invalid. I'm fine with moving the check, perhaps in https://github.com/rancher/rancher/blob/master/pkg/image/export/main.go?

prachidamle: I think the place to add this will be the download link on kontainerdrivermetadata -https://github.com/rancher/rancher/blob/master/pkg/api/customization/kontainerdriver/actionhandler.go#L94

superseb: I put it in here to make this one fail so that the build fails when the images are not correct, versus discovering it when its too late, see https://github.com/rancher/rancher/blob/master/scripts/package#L42

Will putting it there prevent it from being built/released?

kinarashah: @superseb true, but since now the method is used twice - during build + the download images link after running server, I'd think we can get check and panic here: 
https://github.com/rancher/rancher/blob/master/pkg/image/export/main.go#L68 
but leaving the mirror function as it is

superseb: Will change, thanks @prachidamle @kinarashah

prachidamle:
@superseb change here are good, and is there a separate PR for adding the error checks to https://github.com/rancher/rancher/blob/master/pkg/image/export/main.go#L68?

superseb:
@prachidamle @kinarashah  I created a PR for that here https://github.com/rancher/rancher/pull/25021

PR #1068, opened 2020-01-13
cis types for adding info to k-d-m

leodotcloud:
https://github.com/rancher/rancher/issues/24935


apis/management.cattle.io/v3/cluster_scan_types.go
```diff
@@ -78,3 +78,29 @@ type ClusterScan struct {
 	Spec   ClusterScanSpec   `json:"spec"`
 	Status ClusterScanStatus `yaml:"status" json:"status,omitempty"`
 }
+
+type CisBenchmarkVersionInfo struct {
+	MinKubernetesVersion string `yaml:"min_kubernetes_version" json:"minKubernetesVersion"`
```
kinarashah: Do you envision use of MaxKubernetesVersion at some point as well?

leodotcloud: CIS benchmark doesn't have any limitations on max, hence probably not.

apis/management.cattle.io/v3/cluster_scan_types.go
```diff
@@ -78,3 +78,29 @@ type ClusterScan struct {
...
+	BenchmarkVersion string `yaml:"benchmark_version" json:"benchmarkVersion"`
+}
+
+type CisConfig struct {
+	types.Namespaced
+
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	DefaultConfig CisDefaultConfig `yaml:"default_config" json:"defaultConfig,omitempty"`
```
kinarashah: Confused as to why it's named `DefaultConfig`? It could either have BenchmarkVersion directly, or if empty would default to some value?

leodotcloud: At this point, we just have one field `BenchmarkVersion` but we could be adding more, hence created a struct.

leodotcloud: As requested renamed the struct

PR #1069, opened 2020-01-17
Add cluster upgrade strategy to rke

mrajashree:
https://github.com/rancher/rke/issues/1772


apis/management.cattle.io/v3/rke_types.go
```diff
@@ -59,6 +59,14 @@ type RancherKubernetesEngineConfig struct {
 	RotateCertificates *RotateCertificates `yaml:"rotate_certificates,omitempty" json:"rotateCertificates,omitempty"`
 	// DNS Config
 	DNS *DNSConfig `yaml:"dns" json:"dns,omitempty"`
+	// Upgrade Strategy for the cluster
+	UpgradeStrategy *NodeUpgradeStrategy `yaml:"upgrade_strategy,omitempty" json:"upgradeStrategy,omitempty"`
+}
+
+type NodeUpgradeStrategy struct {
+	MaxUnavailable string         `yaml:"max_unavailable" json:"maxUnavailable,omitempty" norman:"default=50%"`
```
superseb: Is there no benefit in having this as int?

mrajashree: I'm accepting a string so that a number and a percentage both can be passed. It gets parsed by the k8s intstr package https://github.com/rancher/rke/pull/1800/files#diff-ee57d1a642f3b4ee28fb4929afe13026R64

apis/management.cattle.io/v3/rke_types.go
```diff
@@ -59,6 +59,15 @@ type RancherKubernetesEngineConfig struct {
 	RotateCertificates *RotateCertificates `yaml:"rotate_certificates,omitempty" json:"rotateCertificates,omitempty"`
 	// DNS Config
 	DNS *DNSConfig `yaml:"dns" json:"dns,omitempty"`
+	// Upgrade Strategy for the cluster
+	UpgradeStrategy *NodeUpgradeStrategy `yaml:"upgrade_strategy,omitempty" json:"upgradeStrategy,omitempty"`
+}
+
+type NodeUpgradeStrategy struct {
+	// MaxUnavailable input can be a number of nodes or a percentage of nodes (example, max_unavailable: 2 OR max_unavailable: 20%)
+	MaxUnavailable string          `yaml:"max_unavailable" json:"maxUnavailable,omitempty" norman:"default=50%"`
```
dramich: Curious where the 50% comes from? That seems high as a default amount of nodes we will take out

mrajashree: Currently RKE upgrades 50 nodes at a time. With the new logic too RKE upgrade will be capped at 50. So with 50% default, lesser nodes will go down during upgrade for a smaller cluster. Whereas for bigger clusters the behavior will be same.

cloudnautique: So 50% seems high. Ultimately, I would say in a three node cluster we want 1 at a time. In larger clusters we probably do not want to invoke such a large change by default. Something like 10% min 1, or something else that accomplishes the first goal, and scales reasonably well in larger clusters. Users will have the ability to adjust the default.

mrajashree: From offline discussion with @kinarashah  reducing it to 10% can cause some problems on the agent side for rancher clusters. And this default will apply to both, rancher and rke. But I'll let @kinarashah comment more on that.
For RKE standalone we should be okay reducing this. But I just wanted to mention again even if we reduce the percentage, RKE will limit parallel upgrades to 50 nodes for large clusters. This number was decided in RKE to avoid running too many goroutines by upgrading too many nodes in parallel. So default of 50% isn't taking all 500 nodes down in a 1000 node cluster

mrajashree:
Based on offline discussion, upgrade strategy will have a default of 10% and min 1

PR #1070, opened 2020-01-22
Dedupe management

dramich:
Problem:
Management context was created for all cluster contexts, this causes
large memory and cpu consumption

Solution:
Only use a single management context for all cluster contexts


ibuildthecloud:
LGTM

dramich:
https://github.com/rancher/rancher/pull/25006

PR #1082, opened 2020-02-07
Shibboleth Auth Provider Types

luthermonson:
Moves LDAP Configs to a new type and pulls it into Shibboleth.

rancher/rancher#24235


luthermonson:
had to redo for build

apis/management.cattle.io/v3/authn_types.go
```diff
@@ -256,36 +256,39 @@ type ActiveDirectoryTestAndApplyInput struct {
...
 
-type LdapConfig struct {
-	metav1.TypeMeta   `json:",inline"`
-	metav1.ObjectMeta `json:"metadata,omitempty"`
-	AuthConfig        `json:",inline" mapstructure:",squash"`
-
-	Servers                         []string `json:"servers,omitempty"                     norman:"type=array[string],notnullable,required"`
-	Port                            int64    `json:"port,omitempty"                        norman:"default=389,notnullable,required"`
-	TLS                             bool     `json:"tls,omitempty"                         norman:"default=false,notnullable,required"`
+type LdapFields struct {
```
cjellick: This is probably mostly fine, but are you really using ALL these fields when using ldap as the search config for saml?
Some that I question if you are using (maybe your are):
- userDisabledBitMask
- userLoginAttribute

It _probably isn't worth refactoring to drop these fields from the search config, but If we are showing unused fields in the UI, I'd like to hide them.

apis/management.cattle.io/v3/authn_types.go
```diff
@@ -349,6 +352,11 @@ type OKTAConfig struct {
 	SamlConfig `json:",inline" mapstructure:",squash"`
 }
 
+type ShibbolethConfig struct {
+	SamlConfig `json:",inline" mapstructure:",squash"`
+	LdapFields `json:",inline" mapstructure:",squash"`
```
cjellick: These ldap fields should not be inlined. They should be under their own ldapSearchConfig property.
The presence or absence of this optional property should indicate whether the user has configured this feature or not.

apis/management.cattle.io/v3/authn_types.go
cjellick: If you did refactor (I think you are going to have to make another pr to address my below comment), you could put all the common fields in LdapFieldsBase and then add a struct that embeds those and adds fields
Also, this should be captialized as LDAPFields to follow proper go conventions

cjellick: im told LDAPFields will break go generate, so i guess dont change that

PR #1087, opened 2020-02-11
CIS changes: cis-1.5, profiles, schedule, alerts

leodotcloud:


apis/management.cattle.io/v3/cluster_scan_types.go
```diff
@@ -104,3 +130,22 @@ type CisBenchmarkVersion struct {
 
 	Info CisBenchmarkVersionInfo `json:"info" yaml:"info"`
 }
+
+type ScheduledClusterScanConfig struct {
+	// Cron Expression for Schedule
+	Schedule string `yaml:"schedule" json:"schedule,omitempty"`
```
leodotcloud: change to cronSchedule

apis/management.cattle.io/v3/cluster_scan_types.go
```diff
@@ -133,19 +133,19 @@ type CisBenchmarkVersion struct {
 
 type ScheduledClusterScanConfig struct {
 	// Cron Expression for Schedule
-	Schedule string `yaml:"schedule" json:"schedule,omitempty"`
+	CronSchedule string `yaml:"cronSchedule" json:"schedule,omitempty"`
```
kinarashah: json tag also needs to be updated to `cronSchedule`

maggieliu:
Issues:
Schedule/Alert: https://github.com/rancher/rancher/issues/25326
Profiles: https://github.com/rancher/rancher/issues/25325
CIS 1.5 support: https://github.com/rancher/rancher/issues/24987

apis/management.cattle.io/v3/cluster_scan_types.go
```diff
@@ -104,3 +130,22 @@ type CisBenchmarkVersion struct {
...
 }
+
+type ScheduledClusterScanConfig struct {
+	// Cron Expression for Schedule
+	CronSchedule string `yaml:"cronSchedule" json:"schedule,omitempty"`
+	// Number of past scans to keep
+	Retention int `yaml:"retention" json:"retention,omitempty"`
+}
+
+type ScheduledClusterScan struct {
```
kinarashah: why is this not ClusterScan with `scheduled` bool? trying to understand why it needs to be called ScheduledClusterScan

apis/management.cattle.io/v3/alerting_types.go
```diff
@@ -199,6 +200,11 @@ type CommonRuleField struct {
 	TimingField
 }
 
+type ClusterScanRule struct {
+	ScanRunType  ClusterScanRunType `json:"scanType,omitempty" norman:"required,options=manual|scheduled,default=scheduled"`
```
kinarashah: json scanRunType

apis/management.cattle.io/v3/cluster_scan_types.go
leodotcloud: ScheduledClusterScan is the struct used to store the info at the cluster level. ClusterScan object is the one created when an actual scan is run.

apis/management.cattle.io/v3/cluster_scan_types.go
leodotcloud: Thank you, will update.

apis/management.cattle.io/v3/alerting_types.go
leodotcloud: thank you, will update.

apis/management.cattle.io/v3/cluster_scan_types.go
```diff
@@ -104,3 +130,22 @@ type CisBenchmarkVersion struct {
...
+	// Cron Expression for Schedule
+	CronSchedule string `yaml:"cron_schedule" json:"cronSchedule,omitempty"`
+	// Number of past scans to keep
+	Retention int `yaml:"retention" json:"retention,omitempty"`
+}
+
+type ScheduledClusterScan struct {
+	// Enable or disable scheduled scans
+	Enabled        bool                        `yaml:"enabled" json:"enabled,omitempty" norman:"default=false"`
+	ScheduleConfig *ScheduledClusterScanConfig `yaml:"schedule_config" json:"scheduleConfig,omitempty"`
```
kinarashah: this is a bit confusing, since we have both schedule and scan prefixed `ClusterScanConfig`...one's a scheduleConfig and another's a clusterScanConfig

kinarashah: can we have struct names also similar to yaml and json tags? they look correct

apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -103,6 +103,7 @@ type ClusterSpecBase struct {
 	EnableClusterMonitoring              bool                           `json:"enableClusterMonitoring" norman:"default=false"`
 	WindowsPreferedCluster               bool                           `json:"windowsPreferedCluster" norman:"noupdate"`
 	LocalClusterAuthEndpoint             LocalClusterAuthEndpoint       `json:"localClusterAuthEndpoint,omitempty"`
+	ScheduledClusterScan                 *ScheduledClusterScan          `json:"scheduledClusterScan,omitempty"`
```
prachidamle: Maybe "ScanSchedule" makes sense instead of "ScheduledClusterScan"? It will be cluster.scanschedule?

leodotcloud: If we introduce some other kind of scan in future, then we won't be able to able to differentiate. ClusterScan is pretty specific.

apis/management.cattle.io/v3/cluster_scan_types.go
leodotcloud: I used `schedule_config` to make it easier for user to configure things in cluster.yml, but if we use the same for struct names, v3.ScheduleConfig doesn't make much sense. Hence, I ended up with the current state.

PR #1091, opened 2020-02-13
Added two new fields in IngressConfig for http and https ports

ibrokethecloud:
Updated IngressConfig to have additional fields for http and https port spec in rke cluster.yml.

This should allow these fields to be consumed by the kdm ingress template.

[Original Issue](https://github.com/rancher/rke/issues/1876)


ibrokethecloud:
@superseb PR for types related to RKE PR [1911](https://github.com/rancher/rke/pull/1911)

apis/management.cattle.io/v3/rke_types.go
```diff
@@ -421,6 +421,10 @@ type IngressConfig struct {
 	ExtraVolumes []ExtraVolume `yaml:"extra_volumes" json:"extraVolumes,omitempty" norman:"type=array[json]"`
 	// Extra volume mounts
 	ExtraVolumeMounts []ExtraVolumeMount `yaml:"extra_volume_mounts" json:"extraVolumeMounts,omitempty" norman:"type=array[json]"`
+	// Http port for ingress controller daemonset
+	HttpPort string `yaml:"http_port" json:"http_port,omitempty"`
```
mrajashree: I think we should set the current values 80 and 443 as default values for these two fields either here or in rke

ibrokethecloud: I have set up default values in the ingress template in kdm

ibrokethecloud: Not applicable anymore since the types have moved to RKE and I have made the changes after re-basing from upstream.

PR #1095, opened 2020-02-18
Add helm version

brendarearden:
Add helm version to determine if app was deployed by helm 2 or helm 3

### DIFF
--- a/apis/management.cattle.io/v3/multi_cluster_app.go
+++ b/apis/management.cattle.io/v3/multi_cluster_app.go
@@ -40,6 +40,7 @@ type MultiClusterAppSpec struct {
 type MultiClusterAppStatus struct {
 	Conditions   []v3.AppCondition `json:"conditions,omitempty"`
 	RevisionName string            `json:"revisionName,omitempty" norman:"type=reference[multiClusterAppRevision],required"`
+	HelmVersion  string            `json:"helmVersion,omitempty" norman:"nocreate,noupdate"`
 }
 
 type Target struct {
--- a/apis/project.cattle.io/v3/app_types.go
+++ b/apis/project.cattle.io/v3/app_types.go
@@ -45,6 +45,7 @@ type AppStatus struct {
 	Notes                string            `json:"notes,omitempty"`
 	Conditions           []AppCondition    `json:"conditions,omitempty"`
 	LastAppliedTemplates string            `json:"lastAppliedTemplate,omitempty"`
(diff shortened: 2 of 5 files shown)


apis/project.cattle.io/v3/app_types.go
```diff
@@ -29,6 +29,7 @@ type AppSpec struct {
 	Prune               bool              `json:"prune,omitempty"`
 	MultiClusterAppName string            `json:"multiClusterAppName,omitempty" norman:"type=reference[/v3/schemas/multiclusterapp]"`
 	ValuesYaml          string            `json:"valuesYaml,omitempty"`
+	HelmVersion         string            `json:"helmVersion,omitempty"`
```
mrajashree: do we need nocreate/noupdate tag?

brendarearden: yes I believe so  - I will add that! 😄

apis/management.cattle.io/v3/multi_cluster_app.go
```diff
@@ -35,6 +35,7 @@ type MultiClusterAppSpec struct {
 	Roles                []string        `json:"roles,omitempty" norman:"type=array[reference[roleTemplate]],required"`
 	RevisionHistoryLimit int             `json:"revisionHistoryLimit,omitempty" norman:"default=10"`
 	UpgradeStrategy      UpgradeStrategy `json:"upgradeStrategy,omitempty"`
+	HelmVersion          string          `json:"helmVersion,omitempty" norman:"nocreate,noupdate"`
```
dramich: Let's move this to status instead. Spec indicates something that a user sets while status is controlled internally. Also in the future status will be protected similar to how k8s does it but through 'steve'.

apis/project.cattle.io/v3/app_types.go
```diff
@@ -29,6 +29,7 @@ type AppSpec struct {
 	Prune               bool              `json:"prune,omitempty"`
 	MultiClusterAppName string            `json:"multiClusterAppName,omitempty" norman:"type=reference[/v3/schemas/multiclusterapp]"`
 	ValuesYaml          string            `json:"valuesYaml,omitempty"`
+	HelmVersion         string            `json:"helmVersion,omitempty" norman:"nocreate,noupdate"`
```
dramich: Let's move this to status instead. Spec indicates something that a user sets while status is controlled internally. Also in the future status will be protected similar to how k8s does it but through 'steve'.

PR #1098, opened 2020-02-21
Add RancherK3sConfig to Cluster type

daxmc99:
Add type support for k3s clusters to support k3s upgrade work.

Part of: https://github.com/rancher/rancher/issues/25371


cjellick:
LGTM but @daxmc99 can you please break this into two commits?
Please get this merged and into rancher today (even if it is just the types updates) so that @westlywright can start doing UI work for it.

cjellick:
@daxmc99 We should give the users the ability to control the concurrency for master nodes too.

Note: RKE just added fields for maxUnavailableWorker and maxUnavailableMaster. We might want to change these to match those names. But I haven't decided yet. Keep the word concurrency for now (and merge it as concurrency) and we'll have a discussion when I'm back on whether we should flip it to maxUnavailable. The reason I hesitate: I'm not sure if the upgrade-controller fully behaves like RKE's maxUnavailable logic does, so I don't want to give a false sense of behavior similarity.

dramich:
@cjellick @daxmc99  can we drop `Rancher` from the names of things? That seems redundant and doesn't follow existing naming `RancherK3sConfig` > `K3sConfig`

PR #1099, opened 2020-02-21
Bump prometheus operator dependency

daxmc99:
This addresses the prometheus@v2.9.2+incompatible error


dramich:
Can you split into two commits? Vendor and generated

PR #1100, opened 2020-02-23
Add maxUnavailable for controlplane nodes

mrajashree:
https://github.com/rancher/rancher/issues/25599


dramich:
Is there a rancher portion to this or API only at this point?

PR #1101, opened 2020-02-24
Adding k8s Version and Cluster Object to Backup

luthermonson:
Adds two new strings to the EtcdBackup in order to grab and use them during the restore process.
https://github.com/rancher/rancher/issues/22232

### DIFF
--- a/apis/management.cattle.io/v3/backup_types.go
+++ b/apis/management.cattle.io/v3/backup_types.go
@@ -66,6 +66,10 @@ type EtcdBackupSpec struct {
 
 type EtcdBackupStatus struct {
 	Conditions []EtcdBackupCondition `json:"conditions"`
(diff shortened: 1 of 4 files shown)


apis/management.cattle.io/v3/backup_types.go
```diff
@@ -62,6 +62,10 @@ type EtcdBackupSpec struct {
 	Filename string `yaml:"filename" json:"filename,omitempty" norman:"noupdate"`
 	// backupConfig
 	BackupConfig BackupConfig `yaml:",omitempty" json:"backupConfig,omitempty" norman:"noupdate"`
+	// version of k8s in the backup pulled from rke config
```
dramich: Both fields need to be moved to status, spec should only be use for fields intended for a user to set.

PR #1102, opened 2020-02-25
Reverse weavework image mirror

StrongMonkey:


StrongMonkey:
@deniseschannon It turns on that mirror weaveworks image is a little bit harder in automation. I'll reverse the change for now and fix in our mirror script so it can support tagging with manifest.

PR #1119, opened 2020-03-06
Revert "token hashing"

cbron:


cbron:
@rmweir nope, its the same

PR #1136, opened 2020-04-07
Revert system images tags for latest k8s versions

superseb:
K8s versions `v1.14.10-rancher1-1` and `v1.15.11-rancher1-0` don't need any other system images updates besides the k8s image itself


deniseschannon:
@superseb when reviewing the 1.14.8 template, shouldn't the calico controllers be removed form the 1.14.10 template?

PR #1142, opened 2020-05-14
Istios UseSourceIP must be nullable or it fails LBs validation

cbron:
You can have only one ConsistentHashLB setting, and since useSourceIp
was not nullable and defaulting to false, it was always getting
set. Meaning setting any of the either two would fail validation.
This PR turns bool to pointer, making it nullable.
See: https://archive.istio.io/v1.4/docs/reference/config/networking/destination-rule/#LoadBalancerSettings-ConsistentHashLB

Issue: https://github.com/rancher/rancher/issues/25515

See here: /v3/project/schemas/consistentHashLB
```
"useSourceIp": {
  "create": true,
  "default": false,
  "type": "boolean",
  "update": true
}
```
=> 

```
"useSourceIp": {
  "create": true,
  "nullable": true,
  "type": "boolean",
  "update": true
}

### DIFF
--- a/apis/project.cattle.io/v3/schema/schema.go
+++ b/apis/project.cattle.io/v3/schema/schema.go
@@ -1115,6 +1115,9 @@ func istioTypes(schemas *types.Schemas) *types.Schemas {
 		MustImport(&Version, istiov1alpha3.VirtualService{}, projectOverride{}, struct {
 			Status interface{}
 		}{}).
+		MustImport(&Version, istiov1alpha3.ConsistentHashLB{}, struct {
+			UseSourceIP *bool `json:"useSourceIp,omitempty"`
+		}{}).
 		MustImport(&Version, istiov1alpha3.DestinationRule{}, projectOverride{}, struct {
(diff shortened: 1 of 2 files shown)


cbron:
@rmweir so this fixes it, and there is a precedent right above my fix for a different part, but I had originally tried to fix via this and it didn't have the same effect: https://github.com/cbron/pkg/commit/475a2bf0af5da25c438fafec767b17166bae2cf3

There weren't any changes in vendor files between the two approaches, the only difference is that change in schema.go, any idea why ? Or which way is preferred ?

rmweir:
@cbron I think nullable is just whether or not the api layer accepts the value null for the key. which works. Otherwise, if the value were omitted I would think that it would pass validation. Are you sure the UI is not just passing the false value?

rmweir:
@cbron I think the nullable schema is saying that the value in the request can explicitly be set to null, in order for that to be true the struct must also support null. However, if the schema is not nullable, you could still represent the value as null by omitting it.

PR #1145, opened 2020-05-18
Add UserInfo field to UserAttributes

mrajashree:
https://github.com/rancher/rancher/issues/27028


apis/management.cattle.io/v3/authn_types.go
```diff
@@ -80,6 +80,7 @@ type UserAttribute struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	UserName        string
+	UserInfo        map[string]string
```
prachidamle, line 83: We could store the the userPrincipal v3.Principal directly here instead of another map[string]string. Also if there is a need to store attributes for more than one auth provider in future, storing the principal per provider will be helpful.

PR #1166, opened 2020-06-22
add tolerations to common logging field

paynejacob:
For: https://github.com/rancher/rancher/issues/20565

Added tolerations to all logging configurations objects.

### DIFF
--- a/apis/management.cattle.io/v3/logging_types.go
+++ b/apis/management.cattle.io/v3/logging_types.go
@@ -52,6 +52,7 @@ type LoggingCommonField struct {
 	OutputFlushInterval int               `json:"outputFlushInterval,omitempty" norman:"default=60"`
 	OutputTags          map[string]string `json:"outputTags,omitempty"`
 	EnableJSONParsing   bool              `json:"enableJSONParsing,omitempty"`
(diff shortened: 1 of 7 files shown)


apis/management.cattle.io/v3/logging_types.go
```diff
@@ -52,6 +52,7 @@ type LoggingCommonField struct {
 	OutputFlushInterval int               `json:"outputFlushInterval,omitempty" norman:"default=60"`
 	OutputTags          map[string]string `json:"outputTags,omitempty"`
 	EnableJSONParsing   bool              `json:"enableJSONParsing,omitempty"`
+	Tolerations			*[]v1.Toleration	  `json:"tolerations,omitempty"`
```
cbron: * Make sure gofmt is enabled/run, you can start a watcher with Goland. 
* Any reason to make it a pointer ? Typically we do that to say null or not, but here its an array so we can just check if empty.

paynejacob: I saw it as a pointer in other locations but for our use case an empty array works better.

cbron:
lgtm, but can you squash your 3 commits into one

paynejacob:
This functionality exists in logging v2 on rancher 2.5.0. This version of logging is deprecated.

PR #1170, opened 2020-07-08
EKS Config types

rmweir:
**Issue:**
Relates to EKSv2 integration and should facilitate:
https://github.com/rancher/rancher/issues/27739


cbron (commented):
Seems fine, small thing. This is just adding right ? No modifications of existing types.

rmweir:
> Seems fine, small thing. This is just adding right ? No modifications of existing types.



> Seems fine, small thing. This is just adding right ? No modifications of existing types.

Correct, it is adding a field to cluster type but it is not taking anything away and should not interfere with any current behavior on its own.

apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/gob"
 	"strings"
 
+	v12 "github.com/rancher/eks-controller/pkg/apis/ke.cattle.io/v1"
```
cbron: Should have a more unique name than `v12`, something relating to eks

apis/management.cattle.io/v3/cluster_types.go
```diff
@@ -329,3 +332,10 @@ type SaveAsTemplateOutput struct {
 	ClusterTemplateName         string `json:"clusterTemplateName,omitempty"`
 	ClusterTemplateRevisionName string `json:"clusterTemplateRevisionName,omitempty"`
 }
+
+type EKSStatus struct {
+	GeneratedVirtualNetwork string   `json:"generatedVirtualNetwork"`
+	GeneratedSubnets        []string `json:"generatedSubnets"`
+	GeneratedSecurityGroups []string `json:"generatedSecurityGroups"`
+	GeneratedServiceRole    string   `json:"generatedServiceRole"`
+}
```
prachidamle, line 341: Wondering if this information can be generic and apply to other kind of clusters as well later? So instead of EKSStatus, maybe these fields can be carried together in a generic wrapper like "NetworkStatus" or similar

rmweir: I think these are subject to change and differ between provider, it makes sense to me for them to be in their own struct, if we end up doing something in the future that overlaps then I think we can revisit this. As it is right now, it is very easy for me to know which fields are specific to the EKS implementation and they match a subset of fields that appear in EKSClusterConfig status.

brendarearden (approved):
LGTM

prachidamle (approved):
LGTM

cbron:
@rmweir can we close this and move to rancher

PR #1176, opened 2020-07-27
Revert minio version update

pennyscissors:
**Problem:** 
When the minio workload is deployed during a pipeline run, it will
fall into a crashloop or restart multiple times causing the pipeline
to fail.

**Related Issues:** 
https://github.com/rancher/rancher/issues/28025
https://github.com/rancher/rancher/issues/28031


pennyscissors:
The issue was fixed [here](https://github.com/rancher/rancher/pull/28184) and reverting is no longer necessary

PR #1179, opened 2020-09-05
Add a CanalControllers image field for Canal

Oats87:
Add a CanalControllers image field for Canal as it is needed for Calico/Canal versions higher than v3.14.0

We are bumping Canal to v3.16.0 which requires a calico-controller to update/modify HostEndpoints.

https://github.com/rancher/rancher/issues/26717

Signed-off-by: Chris Kim <oats87g@gmail.com>


kinarashah (approved):

luthermonson:
@Oats87 this is pretty old now and we put types over to rancher/rancher now. probably need to close this and reopen against rancher/rancher if you still need it

PR #1083, opened 2020-02-10
[2.4] [Fix] Updated api bool fields with default=true to *bool, to be usable from client

rawmind0:
This PR updates api bool fields with default=true to *bool, to be usable from client:
- `IgnoreDockerVersion` on `rke_types`
- `UseInternalIPAddress` on `NodeCommonParams` and `IgnoreDaemonSets` on `NodeDrainInput` on `machine_types`
- `Compress` on `FluentForwarderConfig` on `logging_types`
- `TLS` on `SMTPConfig` on `alerting_types`

This is related with golang `Marshal` function from `encoding/json` that omits false bool if `omitempty` is set, https://github.com/golang/go/issues/13284#issuecomment-157288425

Fix issues https://github.com/terraform-providers/terraform-provider-rancher2/issues/41, https://github.com/terraform-providers/terraform-provider-rancher2/issues/264 and https://github.com/rancher/rancher/issues/25321

PR #1183, opened 2020-10-05
updated NodeUpgradeStrategy.Drain to use a pointer

paynejacob:
https://github.com/rancher/rancher/issues/29058
https://github.com/rancher/terraform-provider-rancher2/issues/440

Allow the drain flag on the NodeUpdateStrategy struct to be nil instead of defaulting to false.


superseb (changes requested):
This sounds like https://github.com/rancher/types/pull/1083 but there was no removal of `omitempty` in that PR, can you describe why it is removed for `Drain` and why that is different?

paynejacob:
@superseb I had removed it because I miss read the reqs I added it back in .

superseb (approved):

paynejacob:
closing in favor of https://github.com/rancher/rke/pull/2278

PR #1193, opened 2021-01-13
[release/v2.4] Fix var interpolation 

pennyscissors:
Issue: https://github.com/rancher/rancher/issues/16148
Related: https://github.com/rancher/rancher/issues/30683
As discussed, the logic to map container's environment variables to our own type has been removed in favor of using k8s native types.


apis/project.cattle.io/v3/schema/schema.go
```diff
@@ -529,13 +529,6 @@ func podTypes(schemas *types.Schemas) *types.Schemas {
 			m.Move{From: "readOnlyRootFilesystem", To: "readOnly"},
 			m.Move{From: "runAsUser", To: "uid"},
 		).
-		AddMapperForType(&Version, v1.Container{},
-			m.Move{From: "command", To: "entrypoint"},
-			m.Move{From: "args", To: "command"},
-			mapper.EnvironmentMapper{},
-			&m.Embed{Field: "securityContext"},
-			&m.Embed{Field: "lifecycle"},
-		).
```
dramich: We can't drop this whole block, just the mapper that tied to the envvars

dramich: `mapper.EnvironmentMapper{}`

pennyscissors: I was about to write a comment asking about this. I'll update it, thanks!

paynejacob (approved):

dramich (commented):
The changes look fine but this is a breaking change to the API so we need to check the CLI and terraform provider to see if they need updated as well

PR #1199, opened 2021-06-21
[release/v2.4] Update types for answerSetString support

pennyscissors:
Issue: https://github.com/rancher/rancher/issues/33072


dramich:
If the entire fix it not going into 2.4 we shouldn't make this change and instead update the CLI to use rancher/rancher

pennyscissors:
Closing due to comment above.

